PR #59, opened 2021-06-08
Add Interval library

tigoe:
An Interval library with a JS-like API, using millis()


tigoe:
@ArduinoBot license added

PR #157, opened 2021-07-06
add library

DCisHurt:


njh:
@DCisHurt looks like this repo has now disappeared:
https://github.com/Matrix-Robotics/MatrixUltrasonicSensor

PR #204, opened 2021-07-16
Update repositories.txt

mandulaj:


mandulaj:
@ArduinoBot Linting should be fixed now

PR #244, opened 2021-07-20
support for RV8803 RTC for TinyMegaCore based uCs

dattasaurabh82:


dattasaurabh82:
@ArduinoBot added release tag

PR #306, opened 2021-07-31
adding https://github.com/noah1510/unit-system to library list

noah1510:
This adds https://github.com/noah1510/unit-system to the library list.

unit-system provides types for the SI units and allows conversions of units and combining them.
This can be a huge help for developers who want to reduce bugs caused by passing wrong variables (time instead of length).
Custom units can be created as needed and pull requests with more types are always welcome.


noah1510:
@ArduinoBot the tag was created just now

PR #463, opened 2021-09-09
Added the LithiumPowered Library

LK-Simon:
Library specifically to simplify the use of Lithium Batteries for Arduino and ESP-powered projects (using a Coulomb Counter such as LTC4150)


LK-Simon:
@ArduinoBot - Tag created!

PR #601, opened 2021-10-21
Update repositories.txt

Jackal28:
MCP3304 library which does not require the use of SPI/ serial.
Has three different options: either pulling the raw number 0-4095, pulling the voltage input, or pulling an averaged voltage with the option to select the amount of times it is averaged 0-1000, and adding in one's own delay between each averaged amount.


Jackal28:
Tag has been created as the initial version matching the library.properties folder

Jackal28:
@ArduinoBot

PR #774, opened 2021-12-09
Update repositories.txt

dndubins:


dndubins:
Version code changed to v1.2.1 @ArduinoBot.

dndubins:
Version code changed to v1.2.1 @ArduinoBot

dndubins:
System is not picking up the correct version number @ArduinoBot

per1234:
@dndubins please read my reply above. I already explained the situation.

dndubins:
Thank you. I attempted to fix this problem, Low on coffee this morning. @ArduinoBot

dndubins:
@per1234 your advice is greatly appreciated, thanks!!!

dndubins:
PS: I really like the word "automagically".

dndubins:
Ok @ArduinoBot I think I have it all ironed out! I even added a LICENSE file this time.

dndubins:
Giving a shout-out to @ArduinoBot now, because the new version of my library (1.2.1) hasn't yet erased the shame and humiliation of the previous version's errors, as per the following instruction: "Comment on your pull request here in the arduino/library-registry repository, mentioning @ArduinoBot in the comment. Doing this will cause the automated check to run again." Even though if I wait, the system should pick it up anyway :)

per1234:
> "Comment on your pull request here in the arduino/library-registry repository, mentioning @ArduinoBot in the comment. Doing this will cause the automated check to run again."

This only works during the submission process. Once the PR is merged, the bot completely ignores it. And the bot only checks for compliance with the requirements for admission. There is a completely separate system that handles the actual indexing.

>if I wait, the system should pick it up anyway :)

That is correct. The indexer job runs hourly. So you should see the new release appear in the logs page within the hour. However, it might take longer than that to see it in Library Manager because the index is cached on your local machine, and perhaps there is some additional delay for it to work through the CDN cache in the cloud.

dndubins:
Who are you, per1234, and why are you so amazingly helpful? Thank you!

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

dndubins:
Hi, just following up with my question before. I ran arduino-lint on the latest version of my library on GitHub (QuickStats 1.2.1) and there are no errors. Yay! However, it's been 2 weeks, and the Library Manager still has the older version with a bug in it (1.2.0). I am on the Canadian side which takes more time. I was just wondering if this is worth a pull request or just sit tight? Many thanks! -Dave

per1234:
Hi @dndubins. You can see the reason here in the logs:

http://downloads.arduino.cc/libraries/logs/github.com/dndubins/QuickStats/

```
2021/12/25 15:18:41 Checking out tag: v1.2.1
2021/12/25 15:18:41 Arduino Lint found errors:
Click to expand Arduino Lint report


Linting library in /home/ubuntu/workspace/gitclones/github.com/dndubins/QuickStats
ERROR: library.properties version value v1.2.1 is invalid. See                                                          
       https://arduino.github.io/arduino-cli/latest/library-specification/#libraryproperties-file-format (Rule LP020)   
```

The "version value" the error message refers to is here:
https://github.com/dndubins/QuickStats/blob/v1.2.1/library.properties#L2
```
version=v1.2.1
```

The relevant information from [the `library.properties` specification](https://arduino.github.io/arduino-cli/latest/library-specification/#libraryproperties-file-format):
>version - version of the library. Version should be [semver](https://semver.org/) compliant. 1.2.0 is correct; 1.2 is accepted; r5, 003, 1.1c are invalid

And from the semver specification:

https://semver.org/#is-v123-a-semantic-version

>### Is “v1.2.3” a semantic version?
>No, “v1.2.3” is not a semantic version. However, prefixing a semantic version with a “v” is a common way (in English) to indicate it is a version number. Abbreviating “version” as “v” is often seen with version control. Example: git tag v1.2.3 -m "Release version 1.2.3", in which case “v1.2.3” is a tag name and the semantic version is “1.2.3”.

I see you corrected the `version` field value after the fact, but the Library Manager indexer system only looks at the tagged version of the library, so that fix is not visible to it.
---

My recommendations for fixing this problem:

1. Bump the version in library.properties to `1.2.2`:
   ```
   version=1.2.2
   ```
1. Push a new Git tag or GitHub release that contains that change to the repository.
   You are welcome to name the tag however you like. Even though tags are the release unit used by the Arduino Library Manager system, the tag name itself is completely ignored. All versioning is done using the `version` field in the `library.properties` metadata file. So it is your choice whether you want to use the "v" prefix on your tag name or not.

The non-compliant v1.2.1 release will never be added to Library Manager, but 1.2.2 will be functionally identical, so it really doesn't affect the users.

dndubins:
Thanks again for the extra help and your patience! Much appreciated. I'm still getting the workflow down.

dndubins:
@per1234 That worked swimmingly, thanks so much!

PR #876, opened 2022-01-08
Add Tee Log

dirkx:
Add T-Log; the ability to also sent Serial output to a telnet server, mqtt, syslog or a near real time updating webpage hosted on the esp/arduino.


dirkx:
@ArduinoBot tag laid down.

PR #1108, opened 2022-02-22
update library version to 2.0.0

lozziboy:


per1234:
Hello @lozziboy. After the initial addition of the library, the indexer job periodically checks each of the libraries in the index and picks up any new release/tag that meets the requirements listed here:

You can monitor the indexing of the library releases by checking the library's dedicated indexer [logs](https://github.com/arduino/library-registry/blob/main/FAQ.md#user-content-can-i-check-on-library-releases-being-added-to-library-manager) web page:

http://downloads.arduino.cc/libraries/logs/github.com/lozziboy/arduino-serial-variable-table/

I can see that the 2.0.0 release has been successfully indexed, so all is well:

```
2022/02/23 07:20:59 Checking out tag: 2.0.0
2022/02/23 07:20:59 Release cSerialWatcher:2.0.0 already loaded, skipping
```

per1234:
You're welcome. The delay is a little bit confusing, but since the index is pretty big, and growing all the time, the caching is necessary for the sake of efficiency.

2.0.0 is already available in my Arduino IDE, but not yet 2.0.1:

image

PR #1138, opened 2022-03-03
Create repositories.txt

Arya11111:


njh:
@Arya11111 it looks like https://github.com/DFRobot/DFRobot_RS01 has disappeared?
Do you have any information about why?

PR #1252, opened 2022-03-31
Update repositories.txt

chcbaram:
@ArduinoBot


chcbaram:
@ArduinoBot created tag.

PR #1260, opened 2022-04-05
Add repository for Sensirion STS4x and SF06-LF

psachs:
Add repository links to Arduino drivers for Sensirions
temperature sensor STS4x and liquid flow sensors of the SF06
sensor family.


psachs:
@ArduinoBot missing tag on sts4x repository is now present

PR #1272, opened 2022-04-07
Add HUSB238Driver library

luoluomeng:
HUSB238Driver is a new library that allows ESP32 users to communicate with the new power chip HUSB238 via the I2C bus.


luoluomeng:
@ArduinoBot recheck

luoluomeng:
@ArduinoBot recheck

PR #1273, opened 2022-04-07
Add NST1001 driver library

luoluomeng:
This is a simple driver which allows ESP32 can communicate with the NST1001 temperature sensor.


luoluomeng:
@ArduinoBot recheck

PR #1282, opened 2022-04-10
Update repositories.txt

justind000:


njh:
@justind000 it looks like [steamrocket](https://github.com/u-fire/steamrocket) repo has been deleted.

PR #1365, opened 2022-05-07
Adding SIKTEC-MDSwitch library

shlomohass:
SIKTEC-MDSwitch library for using and attaching callbacks to Multi direction switch with interrupt support


shlomohass:
@ArduinoBot check again please

PR #1373, opened 2022-05-08
Add DHTINT

RobTillaart:
integer only version of the DHTNew library to save footprint on processors with less memory.


RobTillaart:
@ArduinoBot
tag 0.1.0 added

PR #1407, opened 2022-05-18
Update repositories.txt

richard26a:


njh:
@richard26a it looks like the repository has now been deleted?

Thanks!

PR #1409, opened 2022-05-19
Update repositories.txt

richard26a:


per1234 (approved):
Thanks @richard26a!

The system is designed to be as flexible as possible in regards to the URL format to give the library maintainers freedom to make their submissions as they prefer. So the previous URL did work fine, but the new one works just as fine.

PR #1479, opened 2022-06-11
Update repositories.txt

wollewald:


wollewald:
This new procedure to add libraries is really great!

PR #1510, opened 2022-06-20
Update repositories.txt

goldfish4tech:
https://github.com/goldfish4tech/Goldfish4Tech/blob/main/library.properties


goldfish4tech:
> * @ArduinoBot

@ArduinoBot

PR #1596, opened 2022-07-10
Add Blues Wireless Notecard Auxiliary Wi-Fi

zfields:


zfields:
@ArduinoBot I created a release matching the version in library.properties

PR #1788, opened 2022-09-01
Update repositories.txt

ogneyar:
added my library


ogneyar:
@ArduinoBot fixed the error

PR #1852, opened 2022-09-15
ABLIC S-58LM20A Temperature sensor Library

RLL-Blue-Dragon:
- Manufacturer : ABLIC
- Model number : S-50LM20A
- Sensor : Temperature
- Component type : ADC component
- Number of Components : Single
- Supported OS (HAL) : Arduino
- Calculation : Floating-point
- Conversion type : Linear
- Moving average filter : Moving average filter select ( Non,SMA,EMA,WMA )
- Diagnosis : Range (Min to Max)


RLL-Blue-Dragon:
@ArduinoBot

RLL-Blue-Dragon:
@ArduinoBot

PR #1866, opened 2022-09-17
Update repositories.txt

eloquentarduino:


eloquentarduino:
@ArduinoBot added tag

PR #1884, opened 2022-09-19
Update repositories.txt

serenewaffles:
add OctoPrinter, a library for talking to OctoPrint


njh:
@serenewaffles it looks like this repo has now been deleted, do you have any information about why?

serenewaffles:
> @serenewaffles it looks like this repo has now been deleted, do you have any information about why?

I was privating a large number of repositories and accidentally privated this one as well. I've fixed the issue and the library should be available again after the next action run.

PR #1896, opened 2022-09-23
New library marisa-esp32

siara-cc:
Please include marisa-esp32 to library registry!!


siara-cc:
@ArduinoBot I fixed the issue. Please re-run the checks.

PR #1898, opened 2022-09-23
added Arduino-DumbDisplay

trevorwslee:
added https://github.com/trevorwslee/Arduino-DumbDisplay


trevorwslee:
@ArduinoBot issue should be fixed

PR #2028, opened 2022-10-28
Add Arduino-Base32 library

dirkx:
Proposal for a tiny library to handle Base32 decoding per rfc4648 for hardware one time password calculators.


dirkx:
Dear @ArduinoBot issues fixed - could you be so kind to try again ?

PR #2029, opened 2022-10-28
TOTP generator

dirkx:
Add time based one time password generator that accepts the SEEDs in the format typically found in Qr codes (unlike the existing options; that rely on the binary seed; which most tools do not directly expose).


dirkx:
@ArduinoBot apologies - I forgot to tag; can you recheck & a botsnack for your troube

PR #2061, opened 2022-11-06
Added MPU6050_IND library to the list

ujur007:
Library for MPU6050 IMU sensor


ujur007:
I have added the tag @ArduinoBot

PR #2149, opened 2022-11-28
Update repositories.txt

alexmaurer-madis:


alexmaurer-madis:
@ArduinoBot try again please

PR #2390, opened 2023-01-17
Update repositories.txt

Jayedt5:


Jayedt5:
@ArduinoBot

per1234 (commented):
Since this repository does not have a valid library, and since the "test" on the repository name makes me think it is not intended to be a genuine contribution of a library of value to the Arduino community, I will close this PR as invalid.

@Jayedt5 I will continue to monitor this PR, so if I have misinterpreted the situation, feel free to comment with an explanation of what this is and I will reconsider.

PR #2391, opened 2023-01-17
Update repositories.txt

Jayedt5:


Jayedt5:
@ArduinoBot

Jayedt5:
@ArduinoBot

per1234 (commented):
Since this repository does not have a valid library, and since the "test2" on the repository name makes me think it is not intended to be a genuine contribution of a library of value to the Arduino community (similar to https://github.com/arduino/library-registry/pull/2390) I will close this PR as invalid.

@Jayedt5 I will continue to monitor this PR, so if I have misinterpreted the situation, feel free to comment with an explanation of what this is and I will reconsider.

Jayedt5:
@per1234 Hello this was unexpected, you are correct, I have basically no idea what I am doing, this was a test for a later library I will personally need, as I need a lot of code on a different machine that I cannot download to

PR #2407, opened 2023-01-23
Update repositories.txt

Naguissa:
Add Naguissa/uUnixDate library - Tiny and simplified library to convert to and from UNIX timestamps.


Naguissa:
Thaks, I solved the issue, @ArduinoBot

That's also why travis-ci was not triggering....

PR #2489, opened 2023-02-18
Update repositories.txt

mcmchris:
Adding mcm-bl0940-lib library


mcmchris:
I solved the error @ArduinoBot

PR #2505, opened 2023-02-22
Update repositories.txt

imbeacon:
This is a fork of the main repository, which was last updated in 2020. Since we have an SDK based on this client, we decided to continue with this library. We also believe that this library provides a lot of opportunities for people who want to build their own IoT devices.


imbeacon:
@ArduinoBot updated library name.

imbeacon:
@ArduinoBot Can be checked.

PR #2538, opened 2023-03-06
Add USBHostMbed5 to library index.

aentinger:


aentinger:
@ArduinoBot I think I shall have fixed the issues.

PR #2539, opened 2023-03-06
Update repositories.txt

MicroBeaut:
@ArduinoBot
Adding a TCone library to Library Manager.


MicroBeaut:
@ArduinoBot
I have already created a release tag that matches the version value in the library's library.properties file.

PR #2566, opened 2023-03-12
Add `tdslite` MSSQL connector library

mustafakemalgilor:
tdslite is a lightweight, platform-independent MS-TDS implementation written in pure C++11 that can work with just 2kB of SRAM.


mustafakemalgilor:
@ArduinoBot

PR #2576, opened 2023-03-17
Add hsmcpp library to registry

igor-krechetov:
C++ library for creating hierarchical state machines / finite state machines in embedded systems.


igor-krechetov:
@ArduinoBot, added a tag to repository.

PR #2652, opened 2023-04-05
Remove URL

tony-bringardner:
@ArduinoBot

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -1,4 +1,3 @@
-https://github.com/tony-bringardner/NetworkMonitor
 https://github.com/marcobrianza/ClickButton
 https://github.com/lnbits/arduino-nostr/
 https://github.com/melopero/Melopero_Cookie_RP2040_Arduino_Library


tony-bringardner:
When I added my library there were some warnings about defendants and it causes some confusion when installing the library via Library manager.

I could not determine how to make changes so I attempted to remove the library from the registry (thinking I could re-add with the current properties).

Now I seem to be stuck with a fork that can’t be completed or undone and no way to create another form.

Can you help?


PS:  I’m not even slightly competent with GIT :-(

per1234:
Hi @tony-bringardner 
>I could not determine how to make changes so I attempted to remove the library from the registry (thinking I could re-add with the current properties).

You distribute changes to the users by making new releases of the library.

https://downloads.arduino.cc/libraries/logs/github.com/tony-bringardner/NetworkMonitor/

Since there is no need to remove the library, I'll close this pull request. I'll continue to monitor the thread, so feel free to comment here if you have any questions or concerns.

Per

PR #2734, opened 2023-04-19
Update repositories.txt

jaemannyeh:


jaemannyeh:
I created a tag v1.0.0 that matches the version value in the library's library.properties file. @ArduinoBot

PR #2813, opened 2023-05-01
Update repositories.txt

Eugeniusz-Gienek:
I have changed the name of my repository from WindyWanderer to Eugeniusz-Gienek - there is already Github redirect in place, but the correct behaviour would be to update all of the links which use it.


per1234 (commented):
Hello @Eugeniusz-Gienek. I am confirming your request.

Information about that here:

MatteoPologruto (approved):

PR #2823, opened 2023-05-07
Update repositories.txt

schnoog:


schnoog:
@ArduinoBot , I've updated the name

PR #2847, opened 2023-05-14
Update repositories.txt

GhassanYusuf:
i have added the NewServo Library and added the library.properties file to the main directory


GhassanYusuf:
@ArduinoBot

PR #2874, opened 2023-05-19
Include TinyMatrixMath library

m516:
Source: https://github.com/m516/TinyMatrixMath


m516:
Thanks! Now there should be a tagged release

m516:
@ArduinoBot

PR #2877, opened 2023-05-21
Updated The FABLAB Bahrain L298 Library

GhassanYusuf:
@ArduinoBot

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -5881,5 +5881,5 @@ https://github.com/rhrhhrhr/MPC_ruih
 https://github.com/sinricpro/teleport-arduino-esp32-library
 https://github.com/midilab/uClock
 https://github.com/sivar2311/WebMonitor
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Closing as invalid because the URL we already have is the most correct.

GitHub redirects the new URL:
https://github.com/fablabbh/FABLAB-L298-Driver

to the old URL:
https://github.com/fablabbh/Arduino-L298-Motor-Library

So you are proposing to change the URL to an outdated one. We will not accept such a change.

@GhassanYusuf it seems you are trying to accomplish something with all these pull requests, but it is not clear to me what you are trying to accomplish. I guess you think the automated system can somehow magically infer your intent, but it doesn't work that way. Instead a human has to clean up all these pull requests. So all you are doing is wasting my time and also wasting your time because you aren't getting any closer to accomplishing your goal.

If you will make the effort to clearly communicate what you would like to accomplish, I would be happy to assist you with that.

PR #2878, opened 2023-05-21
added a library to communicate with Cohere models through an API

ejri:


ejri:
added version tag
@ArduinoBot

PR #2901, opened 2023-05-28
Removing a new library I added earlier today.

zapta:
After using the package in the Arduino IDE I realized that it has portability issues, e.g. because it uses vprintf. Removing it for now.


per1234 (commented):
Hello @zapta. I am confirming your request.

PR #2927, opened 2023-06-04
Update and rename repositories.txt to M304-lib

mhorimoto:


mhorimoto (commented):
rename is my mistake.
no renamed.

PR #2981, opened 2023-06-18
Update repositories.txt

ripred:
added updates for Smooth Arduino library

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -5955,6 +5955,5 @@ https://github.com/stm32duino/X-NUCLEO-53L8A1
 https://github.com/rei-vilo/PDLS_EXT3_Basic_BWRY
 https://github.com/skyfroger/RPlatform
 https://github.com/koendv/cst816t
-https://github.com/bitbank2/bb_rtc
 https://github.com/ripred/Smooth
-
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Hi @ripred. Your pull request only moves the library from one arbitrary line in the file to another. Since this change serves no purpose, I will close the pull request as invalid.

If you were trying to accomplish something other than a random change to the file, please comment here with a detailed description of what you were hoping to achieve and I'll help you out.

You mention the Smooth library, but you already successfully submitted that library with your previous pull request: https://github.com/arduino/library-registry/pull/2980

It seems you are establishing a pattern of submitting these pull requests:

- https://github.com/arduino/library-registry/pull/2941
- https://github.com/arduino/library-registry/pull/2942
- https://github.com/arduino/library-registry/pull/2561

These pull requests waste time for both of us so I hope we can have some communication to avoid the creation of additional such requests. It is true that the submissions are fully automated, but that is the only automated aspect of this repository. So if you are having some difficulty with the Library Manager system or want some other change made then that will be handled by a human and you need to communicate in a way a human can understand. There is nothing in place to psychically interpret the intent behind PRs that make random changes to the file, nor to act on such changes.

Regards,
Per

PR #3043, opened 2023-07-08
Add tedtoal/monitor_printf

tedtoal:


tedtoal:
Not sure if I need to mention @ArduinoBot because it looks like when I added a release tag it may have triggered re-checking. But I don't know if reviews happen automatically or if I need to request one?

PR #3044, opened 2023-07-09
Update repositories.txt

tedtoal:
Update project to try to make it pass tests for submission


per1234 (commented):
Hi @tedtoal

>try to make it pass tests for submission

The library already passed all the checks and the submission was accepted. The library should be available for installation via Library Manager very soon, if not already (it has already been indexed, but the update takes some time to work its way through the cache).

Closing as duplicate of https://github.com/arduino/library-registry/pull/3043

PR #3114, opened 2023-07-25
Added MAX11643 ADC library

big12boy:


big12boy:
@ArduinoBot added release and tag

PR #3194, opened 2023-08-06
Update repositories.txt

epsilonrt:
remove the library epsilonrt/pimp in order to rename it.


per1234 (commented):
Hi @epsilonrt.

> in order to rename it.

We only allow renaming libraries by following the standardized procedure for doing so that is documented here:

PR #3196, opened 2023-08-07
Update repositories.txt

epsilonrt:


per1234 (commented):
I'm not amused @epsilonrt. I clearly explained to you that you should follow the standard procedure for requesting the name change: https://github.com/arduino/library-registry/pull/3194#pullrequestreview-1564423422. So why did you do a double submission after that?

epsilonrt:
Hi per1234,
Sorry for the problem I created.
The procedure is not clear to me. I modified the https://github.com/epsilonrt/pimp/blob/main/library.properties file but the url no longer points to pimp.
Seeing no solution to the problem, I thought it was necessary to push the new repository.
I confess that I am a little lost in all these procedures.
I clearly made a mistake in the name of my library because I did not know the meaning of the word 'pimp' in English.
I would therefore like the 'pimp' library to no longer appear in your catalog.
THANKS.

PR #3218, opened 2023-08-14
Update repositories.txt

abcdaaaaaaaaa:


abcdaaaaaaaaa:
if you want to publish new version what will be do?

per1234 (commented):
@abcdaaaaaaaaa please don't submit pointless pull requests like this. I guess you are trying to find some hacky way to get the Library Manager system to do something, but that will never work.

If there is something you would like to accomplish, or you are having some problem, then take the time to read the documentation. Most likely the there are instructions for the correct way to accomplish your goal or the explanation of your problem is there:

And when you submit a pull request to accomplish something other than a basic submission, please take the time to write a clear description of what the purpose is. The submission system is full automated (though closely monitored by humans), but all other operations are handled by humans, so communicate with us as you would do with a human!!!

If you still have a question or problem, please post on Arduino Forum. I'm sure to see your post and provide assistance over there:

https://forum.arduino.cc/c/development/libraries/39

PR #3289, opened 2023-09-04
Add Arduino_POSIXStorage to repositories.txt

aliphys:

PR #3300, opened 2023-09-07
[AE-158]  Add Arduino_UnifiedStorage to repositories.txt

aliphys:


aliphys:
The Linter has identified one `WARNING` which we will ignore. See: https://github.com/arduino/library-registry/pull/3289#issuecomment-1705111284
>The rule for library name length (LP010) has increased from 16 to 32. Once https://github.com/arduino/arduino-lint/pull/511 is part of the arduino linter release, this warning will be eliminated.

PR #3329, opened 2023-09-14
Update repositories.txt

simonlmn:


simonlmn:
@ArduinoBot fixed library.properties

PR #3341, opened 2023-09-18
Update repositories.txt

ChatCG3:


per1234 (commented):
Hi @ChatCG3. Thanks for your interest in contributing to the Arduino Library Manager registry.

As the "library" in the name implies, this is a place to submit Arduino libraries for inclusion in the Arduino Library Manager. https://github.com/earlephilhower/arduino-pico-littlefs-plugin is not an Arduino library, so it is not appropriate to submit it here.

So I will close this pull request as invalid. You are welcome to submit any actual Arduino libraries if you like.

PR #3358, opened 2023-09-23
Update repositories.txt

weedmanu:
Delete OLD

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -5812,7 +5812,6 @@ https://github.com/waspinator/AccelStepper
 https://github.com/waspinator/CD74HC4067
 https://github.com/wayoda/LedControl
 https://github.com/weckbach/AstroMech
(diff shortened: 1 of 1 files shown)


weedmanu:
For deleted an OLD librarie

per1234 (commented):
Hi @weedmanu. I see you just attempted to redundantly add the library in https://github.com/arduino/library-registry/pull/3357, and now immediately after that are attempting to remove the registration in this pull request.

I am guessing your intent is to remove the library and then immediately submit it once again. That is not allowed because there is never a valid reason to do that.

What you must do instead is follow the correct procedure for whatever it is you are hoping to accomplish. You can learn the correct procedures by reading the documentation:

If you will provide a detailed description of what you are hoping to accomplish in a comment here on this pull request thread, I will be happy to provide guidance. HINT: a competent developer would have done that in the pull request message.

PR #3383, opened 2023-09-27
Update repositories.txt

helson16:
Add espressif/esp-mqtt


per1234 (commented):
Hello . Thanks for your interest in contributing to the Arduino Library Manager index.

As the error reported by the bot indicates, https://github.com/espressif/esp-mqtt is not an Arduino library. This registry is only for Arduino libraries so we cannot accept your submission.

You are welcome to submit any actual Arduino libraries you would like to see added to the Arduino Library Manager though.

Regards,
Per

PR #3398, opened 2023-09-29
Lion:Bit Library Added

PradeepKotu:
for Lion:Bit S3 STEM DEV Board we have added the Library www.lionbit.cc


PradeepKotu:
Close this for new update

PradeepKotu:
> A problem was found with your submission https://github.com/LionBit-STEM-Board/lionBit-STEM-library
> 
> ❌ **ERROR:** The repository has no tags. You need to create a [release](https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) that matches the `version` value in the library's library.properties file.

@ArduinoBot Updated the issue

PR #3449, opened 2023-10-11
Update repositories.txt

F-Schmidt99:
Added URL to a new library that we will use for an Arduino-based high-altitude weather balloon.


F-Schmidt99:
@ArduinoBot

F-Schmidt99:
@ArduinoBot

PR #3489, opened 2023-10-21
updated library url

Tenveis:
updated the name of library in accordance with arduino lint suggestion.


per1234 (commented):
Hi @Tenveis. I am confirming your request.

Unlike submissions, this type of request is still handled manually. The Library Manager index maintainer occasionally comes through to clear all the outstanding requests. They will close this issue when it is done. You will get a notification from GitHub at that time.

Information about that here:

MatteoPologruto (approved):
Done:
```
2023/10/25 16:07:32 Loaded 6304 libraries from DB
Removing SimpleMotorControl
Success!
```

PR #3502, opened 2023-10-24
Cambios 1

AlfredoMunguia:
Primeros cambios


per1234 (commented):
Closing as duplicate of https://github.com/arduino/library-registry/pull/3484

@AlfredoMunguia since you didn't bother to write the reason, it is not clear what you hoped to accomplish by redundantly submitting your library. I can tell you that there is never a case where doing such a thing is productive. 

Please carefully read the documentation to understand the correct way to use the Arduino Library Manager and Library Registry systems:

I think that will allow you to understand how to achieve your goal. If not, you can reply here with a detailed explanation of:

- What you want to accomplish
- What you tried
- What problems you encountered when you tried those things

and I'll provide assistance.

PR #3516, opened 2023-10-25
Update repositories.txt

AlfredoMunguia:
cambios

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -1,4 +1,5 @@
 https://github.com/makerlabvn/Makerlabvn_SimpleMotor
+https://github.com/AlfredoMunguia/TinyGuixhe
 https://github.com/Khuuxuanngoc/DHT-sensor-library
 https://github.com/satspares/DWIN_DGUS_HMI
 https://github.com/CIRCUITSTATE/CSE_ModbusRTU


per1234 (commented):
Hello @AlfredoMunguia. This is the second of these invalid pull requests you have submitted (https://github.com/arduino/library-registry/pull/3502).

By the mention of "cambios" it seems that maybe you are hoping to get the bot to magically recognize some mysterious change you made to your library. It doesn't work like that. If you want to distribute a change to the users, you only need to make a new release of the library. This is done by the following procedure:

1. Make a new [release](https://docs.github.com/en/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the library repository.

---

If you were hoping to accomplish something different, then instead of wasting everyone's time with these pointless pull requests, read the documentation we provide to understand the correct way to accomplish it:

- https://github.com/arduino/library-registry#readme
- https://github.com/arduino/library-registry/blob/main/FAQ.md#arduino-library-manager-faq

If you don't find the answer there, then you can reply here on this thread with a _detailed_ description of what you would like to accomplish, what you tried, and what happened when you tried those things. You are welcome to write that detailed description in your native language and then use an automatic translation service like [**Google Translate**](https://translate.google.com/) to convert it to English before posting it here on GitHub.

PR #3529, opened 2023-10-28
remove url for Library Index

hasenradball:
# Remove url for Library Index https://github.com/affanhanifathtarech/DS3231_RTC

Removing this entry has the following reasons:

1. The library was copied without adapting the metadata correctly in the `library.properties` file
* url points to wrong location
* maintainer is wrong
* include name differs from Library name which causes issues during build process if both libraries are installed.
* => this causes confusion by the users, see PR of maintainer of original Lib [PR](https://github.com/affanhanifathtarech/DS3231_RTC/pull/1)

2. The library was not maintained about Aug 2022, over one year

3. owner of this Library did not respond since Mar 2022 to this PR mentioned above, regarding to correct the entries in the `library.properties` file.

This is a friendly request to remove this library from the index to give the owner a hint to adapt the `library.properties` file accordingly.
If there is no response from him following, I would see this as hint this library is not used anymore in Arduino Index.
You can also see the issue if you search for the `DS3231` library in Arduino IDE which results in two entries pointing to the identical github location (when clicked on `More Info`) but meaning should be two different libraries.

Library `DS3231` and  `DS3231_RTC` points to the same URL when use the `More Info`-button.

Best regards
Frank

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -2626,7 +2626,6 @@ https://github.com/adnan-elabdullah/KmeStepper
 https://github.com/adrian200223/Simple5641AS
 https://github.com/adrien-legrand/as5x47
 https://github.com/aelse/ArduinoStatsd
-https://github.com/affanhanifathtarech/DS3231_RTC
 https://github.com/afpineda/MeanAndVarOnTheFly-Arduino
 https://github.com/agdl/Base64
 https://github.com/agdl/GoPRO


per1234 (commented):
Thanks for taking the time to submit a pull request @hasenradball.

I do appreciate your care for the quality of the Library Manager libraries. However, Arduino's policy is that we only remove libraries under one of the following conditions:

- The maintainer of the library has approved the removal
- There is conclusive proof that the library contains illegal material (e.g., copyright violation)
- There is conclusive proof of malicious content

Since none of these conditions have been met, I will have to decline this. Arduino is committed to distributing the community's libraries, but micromanaging >6000 libraries is not feasible.

@affanhanifathtarech please work with @hasenradball and the other contributors to your project to improve the quality of the library. I know it can be difficult to find time to work on hobby/volunteer projects, but the effort to evaluate and accept the important changes they propose is quite minimal.

PR #3537, opened 2023-10-30
Update repositories.txt

gemi254:


per1234 (commented):
Hi @gemi254. As the bot commented above, you already added this URL to the list a few days ago in https://github.com/arduino/library-registry/pull/3525. So I'll close this as a duplicate.

I'll continue to monitor this thread, so if you hoped to accomplish something other than a redundant submission through this pull request, just comment here with a detailed description of your goal and I'll provide assistance.

PR #3560, opened 2023-11-05
Update repositories.txt

pg-goose:


pg-goose:
@ArduinoBot

pg-goose:
@ArduinoBot

PR #3577, opened 2023-11-08
Update repositories.txt

janikheiler:
i am about to change my username to janikheiler. therefore i would like to change the url :)

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -143,7 +143,7 @@ https://github.com/soylentOrange/DigiSpark_PWM
 https://github.com/SanteriLindfors/WiFiProvisioner
 https://github.com/soylentOrange/Forced-BMX280
 https://github.com/vishesh-varma/DWIN
-https://github.com/whoelsethanjesus/Mintrix
+https://github.com/janikheiler/Mintrix
 https://github.com/DigitalCodesign/Escornabot-Library
 https://github.com/SCRN92/Sim800L
 https://github.com/manuelbl/usb-pd-arduino


per1234 (commented):
Hi @whoelsethanjesus. Thanks for your care about keeping the registry data up to date. I see that the change has not yet occurred, which is the cause of the problem reported by the bot that the new URL https://github.com/janikheiler/Mintrix does not load.

We should wait on making this change until the URL is valid. So you should proceed with changing your GitHub username. Once you have done that, please comment here on the pull request thread to let us know we can proceed.

If you use GitHub's username change feature (instead of creating a new account under the new name), then GitHub automatically sets up redirects from the old URLs of your repositories to the new URLs so you won't break anything by changing your username. However, we still encourage library maintainers to submit URL update requests here even when a redirect is in place because we would prefer to avoid a reliance on those redirects, which aren't guaranteed to continue working forever.

janikheiler:
Hi @per1234 ,
thanks for the comment. I thought I'd wait until you changed it so you have proof that really I wanted to change it and not some other account. Anyways. I changed my name now. :)

per1234 (commented):
Hello @janikheiler. I am confirming your request.

Information about that here:

> I thought I'd wait until you changed it so you have proof that really I wanted to change it and not some other account.

That is very much appreciated. We do verify the legitimacy of requests for changes to existing library registrations. In this particular case, GitHub's redirect of the previous URL to the new URL is sufficient verification, but in other cases it can be challenging to verify requests so you definitely took the appropriate approach to this.

PR #3617, opened 2023-11-18
Update repositories.txt

eb1992:
This Arduino library provides an easy-to-use interface for controlling an LCD display with a shift register. It can control the display with only four pins from the Arduino.


eb1992:
Dear @ArduinoBot , please check again.

PR #3695, opened 2023-12-04
add tmc7300 library

joshua-8:
Trinamic uart controlled dual DC motor controller IC


joshua-8:
Thanks for catching that @ArduinoBot! I tried to fix it

PR #3730, opened 2023-12-11
Update repositories.txt

rogerjames99:
Add entry for json2asw

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -1,3 +1,4 @@
+https://github.com/rogerjames99/json2asw
 https://github.com/littleBitsman/Asynchrony
 https://github.com/alireza7575/MKS_SERVO57
 https://github.com/amiga68k/electuno


rogerjames99:
@ArduinoBot

per1234 (commented):
Hello @rogerjames99. Thanks for your interest in contributing to the Arduino Library Manager index.

```text
ERROR: Path does not contain a valid Arduino library.                         
       See: https://arduino.github.io/arduino-cli/latest/library-specification
       (Rule LS001)  
```

Arduino libraries must contain a header file in the root of the source folder. The current structure you are using, where the header files are all under the `include` subfolder of the library is not a valid Arduino library structure.

Please fix the problem and then test to make sure it works using one of the official Arduino development tools:

- [**Arduino CLI**](https://arduino.github.io/arduino-cli/latest/)
- [**Arduino IDE**](https://docs.arduino.cc/software/ide-v2/tutorials/getting-started/ide-v2-downloading-and-installing)
- [**Arduino Cloud**](https://docs.arduino.cc/arduino-cloud/guides/overview)

Since you do not use these tools yourself, there is a high risk that a future change to your library will break its compatibility with the Arduino development tools. For this reason, I recommend you to set up a simple "smoke test" by adding a GitHub Actions workflow to your repository that compiles an Arduino sketch using the library. Arduino provides a GitHub Actions action that makes this easy to set up:

https://github.com/arduino/compile-sketches#readme

---

You might consider setting up an automated check in your library repository for compliance with the Library Manager requirements. This could run on every commit and pull request to ensure such violations are caught immediately.

---

1. Correct the structure of your library.

If you have any questions or concerns, please let me know. If you decide you no longer wish to submit your library, you can close this pull request.

Regards, Per

PR #3733, opened 2023-12-12
ESP32QuantumAlgorithm

Manojshyaam:


per1234 (commented):
Hello @Manojshyaam. Thanks for your interest in contributing to the Arduino Library Manager index.

> ❌ ERROR: The repository has no tags. You need to create a [release](https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) that matches the version value in the library's library.properties file.

I looked at your repository and see that there is an additional problem, which is that you added the library to the repository as a ZIP file. The repository must contain the library files in order to be allowed in the Arduino Library Manager. It is also just a terrible idea in general to store your code in a ZIP file because you lose all the benefits of a source code repository. GitHub even automatically generates a ZIP file from the source code files in a repository to make it easy for people to download, so your approach is completely unnecessary.

Please fix the problem so we can proceed with your submission. You should carefully study the full list of requirements for inclusion in the Arduino Library Manager to make sure your library is compliant:

1. Bring your library repository into full compliance with the requirements.

If you have any questions or concerns, please let me know. I'm happy to help you to achieve a successful library submission.

Regards, Per

Manojshyaam:
@ArduinoBot

Manojshyaam:
@ArduinoBot

Manojshyaam:
@ArduinoBot

per1234 (commented):
@Manojshyaam I took a look at your code and I see this is not a library. As the name indicates, Arduino Library Manager is only for Arduino libraries so we will not be able to accept this.

PR #3803, opened 2023-12-27
Removal of YH-Choi-001/Kojay library

YH-Choi-001:
May I kindly ask to remove this Kojay library from the Arduino libraries registry, since I would like to keep part of my code hidden from the public as secrets of my organisation.


per1234 (commented):
Hello @YH-Choi-001. I am confirming your request.

PR #3822, opened 2023-12-30
+tflm_esp32,+tflm_cortexm,-EloquentTensorFlowX

eloquentarduino:


per1234 (changes requested):
Hi @eloquentarduino. There are two distinct operations here:

- Submit "tflm_esp32" and "tflm_cortexm" libraries
- Remove "EloquentTensorFlow32" and "EloquentTensorFlowCortexM" libraries

Please submit a separate pull request for each of those operations. The reason I am asking you to do this is that we have a completely automated system for handling submissions, but removal requests are handled manually. The submission handling system only works on pull requests that consist exclusively of submissions (it is fine to submit multiple libraries in a single PR). So if you split this into two pull requests then the PR that consists purely of submissions will be accepted immediately and the PR that consists purely of removal requests will be easier for the maintainers to manage.

PR #3828, opened 2023-12-31
Update repositories.txt with karol-brejna-i/RemoteDebug

karol-brejna-i:


karol-brejna-i:
@ArduinoBot would you mind rechecking the PR?

PR #3842, opened 2024-01-04
Update Documentation

m-shintaro:
Documentation has been updated.
@ArduinoBot


m-shintaro:
@ArduinoBot

per1234 (commented):
Hello @m-shintaro. The way to distribute changes to the users of your library is to make a new [release](https://docs.github.com/en/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the library repository.

This means there is no need to take any additional action to add new releases of your library to Library Manager. They will be added automagically! So I will close this pull request as not needed.

PR #3868, opened 2024-01-08
Update repositories.txt

roshanraj2:


roshanraj2:
arduino/library-registry @ArduinoBot

PR #3871, opened 2024-01-09
Update repositories.txt

eloquentarduino:


cmaglie:
Hi @eloquentarduino,

the problem is that these symbolic links are broken in your library:

https://github.com/eloquentarduino/eloquent_edgeimpulse/blob/1.0.0/src/eloquent_edgeimpulse/circular_buffer.h

they point to some absolute paths similar to this one that are missing:

```
circular_buffer.h -> /Users/simone/Documents/Arduino/libraries/eloquent_commons/circular_buffer.h
```

Please remove the broken links from the library and resubmit it.

PR #3898, opened 2024-01-15
Update repositories.txt trigger update to EdgieD entry

crunchysteve:
Advance EdgieD to version 1.0.0


crunchysteve:
Seriously, I don't understand this stuff. I give up.

PR #3935, opened 2024-01-21
removing ConfigPortal8266 and ConfigPortal32

yhur:
Temporary removal of ConfigPortal8266 and ConfigPortal32 for a  further refinement.


per1234 (commented):
Hi @yhur. It is not necessary to temporarily remove libraries to make refinements. Just make a new release of the libraries to distribute the refinements, as is done by the other thousands of library authors.

You can learn how to make a new release here:

yhur:
@per1234 
Thanks for the explanation.

PR #3939, opened 2024-01-22
Update repositories.txt

xreef:
LoRa EBYTE E70 CC1310 device library tested with Arduino, esp8266, esp32, STM32 and Raspberry Pi Pico (rp2040 boards)..


xreef:
I try to di the fix @ArduinoBot

PR #4006, opened 2024-02-06
Update repositories.txt

aakfcnn:


per1234 (commented):
As the bot explained:

>A problem was found with your submission https://github.com/arduino-libraries/Servo
>
>❌ ERROR: Submission URL is already in the Library Manager index.

So I will close this as duplicate/invalid.

@aakfcnn if you were hoping to accomplish something other than redundantly submitting the "Servo" library, which has been in Arduino Library Manager from day one, you can add a detailed description of what you want to achieve here on the pull request thread.

I'll continue to monitor this thread so I'll be sure to notice it and to provide assistance.

PR #4042, opened 2024-02-11
Update repositories.txt

ajas19:


per1234 (commented):
Hello @ajas19. Thanks for your interest in contributing to the Arduino Library Manager index.

I see that your https://github.com/ajas19/calculator19 project is an Arduino sketch, not a library. As the name implies, Arduino Library Manager is for libraries, not sketches, so we can not accept this submission.

If you would like to share your sketch with the Arduino community, you might consider submitting it as a project to the Arduino Project Hub:

PR #4076, opened 2024-02-18
duplicate library name

koendv:
Renamed FloatToString to FloatToAscii to avoid collision with https://github.com/tedtoal/floatToString


koendv:
@ArduinoBot

PR #4078, opened 2024-02-18
Update repositories.txt to add XY6020L lib

Jens3382:
dcdc converter can be controlled via UART


Jens3382:
@ArduinoBot repo is now public

PR #4146, opened 2024-03-01
New library

cotestatnt:
Removal of library YA_FSM  which has been replaced by the more efficient library AgileStateMachine


cotestatnt:
@ArduinoBot 

Why release with tag 1.0.4 is not loaded?

PR #4152, opened 2024-03-02
Add PowerFeather-SDK library

powerfeatherdev:
PowerFeather-SDK is a library that enables users to access the power management and monitoring features of [ESP32-S3 PowerFeather](https://powerfeather.dev/).


powerfeatherdev:
@ArduinoBot

PR #4171, opened 2024-03-06
Update repositories.txt

ozerngg:
Update repositories


per1234 (commented):
Hello @ozerngg. Thanks for your interest in contributing to the Arduino Library Manager index.

I see that your repository is not a library, but instead a collection of sketches. As the name implies, Arduino Library Manager is only for libraries, not for sketches. So we will not accept this submission.

If you would like to share your projects with the Arduino community, you can consider writing about them and publishing them on the Arduino Project Hub:

PR #4174, opened 2024-03-06
Update repositories.txt

RobTillaart:
Arduino library to implement float16ext data type.
- does not support INF/ NAN 
- uses the full range of internal values -131007..+131007


RobTillaart:
@ArduinoBot
Sorry I was too fast, now the repo is tagged and released.

PR #4266, opened 2024-03-27
 Update repositories.txt

AS-dbrusamento:


AS-dbrusamento:
Lint errors should be resolved @ArduinoBot

PR #4272, opened 2024-03-27
Update repositories.txt

orkungedik:


per1234 (commented):
Closing as completely pointless since the pull request doesn't make any changes.

@orkungedik I guess you hoped to accomplish something other than wasting my time with this pull request, but I don't know what that was because you didn't make the effort to communicate the purpose.

It seems you might be under the impression that this repository is maintained by an omniscient robot that can psychically determine what you are attempting to accomplish by submitting this PR. That is not the case here. Although we have automated the library submission process, all other maintenance of this repository is done by humans. Please act accordingly; as you would do if interacting with a human in person.

We are here to help you, but we can only do that if you make the effort to communicate in a way that a human can understand. Please add a comment here that provides a clear and detailed description of what you are trying to accomplish. I'll be happy to provide assistance once I understand what is needed.

PR #4277, opened 2024-03-28
Update repositories.txt

orkungedik:
sentence variable has updated in library.properties file


per1234 (commented):
Hello @orkungedik. After the initial addition of the library, the indexer job periodically checks each of the libraries in the index and picks up any new release/tag that meets the requirements listed here:

This means that the correct way to distribute new versions of your library is to make a new release of the library. There is no need to take any action in this repository to add new releases of your library to Library Manager. They will be added automagically!

PR #4319, opened 2024-04-06
Update repositories.txt

1ux:


1ux:
I have fixed the lint warnings in the BasicColorLedControl library and renamed it to LED_RGB_Control. 
I have therefore deleted the corresponding old link to the library.

per1234 (commented):
Hi @1ux.

> renamed it to LED_RGB_Control

I see you have changed the value of the `name` field of the `library.properties` metadata field:

https://github.com/1ux/LED_RGB_Control/blob/main/library.properties#L1

```text
name=LED_RGB_Control
```

However, Library Manager works exclusively from Git tags, so the indexer system can't recognize the name change until you create a new [release](https://docs.github.com/en/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the library repository.

If you look at the repository contents at the latest tag, you can see that it still has the old name value:

https://github.com/1ux/LED_RGB_Control/blob/v1.1/library.properties#L1

```text
name=BasicColorLedControl
```

So if you want to request a change of the registered library name, you must first make a new release. Once you have done that, comment here to let me know I can proceed with your request..

Regards, Per

per1234 (commented):
Since there was no response regarding my previous comment about the change to the library name even after some days, I decided to just proceed with the URL update.

PR #4328, opened 2024-04-08
Update repositories.txt

MariaNegrelli:


per1234 (commented):
Hello @MariaNegrelli. Thanks for your interest in contributing to the Arduino Library Manager index.

Unfortunately it is not possible to add this repository. The reason is because is is a collection of many different Arduino libraries in subfolders of a single repository. It is required to for a library to have a dedicated library in order for it to be accepted to Library Manager:

https://github.com/jrowberg/i2cdevlib/issues/311

If you want to submit the library, you would need to create a dedicated repository to host a fork of the library.

PR #4331, opened 2024-04-09
Update repositories.txt to add Arduino_Cellular

cristidragomir97:
Added https://github.com/arduino-libraries/Arduino_Cellular


sebromero:
@ArduinoBot please take another look

PR #4339, opened 2024-04-11
Update repositories.txt

GyverLibs:
I want to re-add these libs in case of wrong versioning and missing versions in PIO registry

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -2422,7 +2422,6 @@ https://github.com/GyverLibs/fixed
 https://github.com/GyverLibs/Forecaster
 https://github.com/GyverLibs/GParser
 https://github.com/GyverLibs/GRGB
-https://github.com/GyverLibs/GSON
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Hi @GyverLibs.

> I want to re-add these libs in case of wrong versioning

We don't allow "re-adding" of libraries. I don't know what you mean by "wrong versioning". If there is a serious problem with a release, you can submit a request to have that specific release removed:

Since release removals are a somewhat labor intensive manual process, please only request removals when a release absolutely must not be offered by Library Manager due to a serious problem. If it is only a bug in the release, simply bumping the version and making a new release to distribute the fix will be sufficient.

> missing versions in PIO registry

This repository is not used in any way by PlatformIO so operations made here won't have any effect on the PlatformIO registry. You should consult the PlatformIO documentation to learn how to address any problems with your libraries related to their registry.

GyverLibs:
ok thanks

PR #4340, opened 2024-04-11
New Version for AtrivaTECH_PicUNO

atulravi:
Version 1.0.1. Also minor url change.


per1234 (commented):
Hi @atulravi.

> Version 1.0.1.

> Also minor url change.

The pull request doesn't make any change at all and it seems the library repository is still at the original URL https://github.com/AtrivaTECH/AtrivaTECH_PicUNO. So I'm going to go ahead and close this pull request as pointless.

If you do want to change the registered library repository URL, you can submit a new pull request.

PR #4354, opened 2024-04-14
Add library for IP2368 QC/PD charging control IC

D-314:


D-314:
@ArduinoBot, 1.0.1 hotfix for wire dependency

PR #4446, opened 2024-04-27
Update repositories.txt

brincode:
Update fa3el_kher


per1234 (commented):
@brincode please don't submit pull requests over and over again. You will never accomplish anything by doing this other than wasting your time and my time. The same result will occur every time. If you didn't get the result you hoped for the first time, then the way to proceed is to carefully study the information provided by the bot as well as the extensive documentation in the readme and the documents linked to from there:

Most likely that will allow you to understand the correct way to proceed. If not, you can add a comment to the pull request thread with a detailed explanation of what you are hoping to accomplish and what it is that wasn't clear in message from the bot and/or the documentation. I monitor all the pull request threads closely so I will see your comment and respond.

Although we have automated the library submission process, all other maintenance of this repository is done by humans. It seems you might be under the impression that this repository is maintained by an omniscient robot that can psychically determine what you are attempting to accomplish by submitting this PR. I am here to help, but I can only do that if you make the effort to communicate with me as a human.

PR #4447, opened 2024-04-27
Update repositories.txt

brincode:


per1234 (commented):
Closing as invalid. The library was already submitted in https://github.com/arduino/library-registry/pull/4442 and this appears to be yet another of many attempts to "Update fa3el_kher", which I will support in https://github.com/arduino/library-registry/pull/4446

PR #4458, opened 2024-04-29
Update repositories.txt

brincode:


per1234 (commented):
Duplicate of https://github.com/arduino/library-registry/pull/4442

Although this would normally be interpreted as an incompetent attempt to request a URL update (the added URL is the renamed `brincode/Fa3el_kher` repo, which is already in the list), and support provided in following the correct procedure to update the URL, the PR author's history of behavior in this repository and contents of the library prove that they are not acting in good faith and so the library will instead be removed from the registry entirely (https://github.com/arduino/library-registry/pull/4462).

PR #4484, opened 2024-05-01
Update repositories.txt

Pascal2511:

PR #4487, opened 2024-05-02
Update repositories.txt

Pascal2511:


Pascal2511:
@ArduinoBot

per1234 (commented):
Closing as duplicate of https://github.com/arduino/library-registry/pull/4484

@Pascal2511 your library has already been added. There is no need to add it again. It does take a little bit of time for the submission to be processed, and then for the updated index to propagate through the cache so that the library is available for installation via Library Manager to all users, but it will get there soon. We give a conservative estimate of universal availability within a day's time.

I can already see it in Library Manager of my own Arduino IDE:

image

I'll continue to monitor this pull request thread, so if you have any questions or concerns you can add a comment here and I'll provide further assistance.

PR #4535, opened 2024-05-18
Added new libraries

Narwhalsss360:


Narwhalsss360:
Have to create a PR for independent ones first, then dependent

PR #4547, opened 2024-05-19
Added dependent libraries

Narwhalsss360:
Since i cant put libraries that depend on each other in one PR, this is the second one.


Narwhalsss360:
@ArduinoBot fixed tag target branch

PR #4563, opened 2024-05-22
Update repositories.txt

ducthaiiot:


per1234 (commented):
As explained by the bot, the "Servo" library is already in Arduino Library Manager so the change you propose here doesn't serve any meaningful purpose. We only accept changes to the contents of this repository when there is a valid reason for doing them so I will close this as invalid.

PR #4565, opened 2024-05-23
Added new malware reporting paragraphs

rhpco:
Paragraph aligned with the internal documents, further details on internal task PNX-4521
cc @per1234


per1234 (approved):
Thanks @rhpco!

PR #4577, opened 2024-05-27
Update repositories.txt

kucingkuco:


per1234 (commented):
Hello @kucingkuco. Thanks for your interest in contributing to the Arduino Library Manager index.

I see that your project is a program rather than a library. As the name implies, the Arduino Library Manager is a system for distributing libraries. So it is not appropriate to submit programs here.

If you would like to share programs with the Arduino community, you might consider doing a write up on the Arduino Project Hub:

PR #4657, opened 2024-06-12
Update repositories.txt

fabianhugo:
- adds Calliope Arduino library


fabianhugo:
- added tag
@ArduinoBot

PR #4684, opened 2024-06-18
Update repositories.txt

HackerNowful:
## Library Name
sd-ducky-v1

## Repository URL
https://github.com/HackerNowful/SD-Ducky

## Release Notes
- Updated to version 1.0.1
- Added support for HID attack with SD card


per1234 (commented):
@HackerNowful it seems you think you must submit an empty pull request every time you make a new release of your library. This is not at all necessary.

This is all explained in the documentation. Please take the time to read it:

Please don't submit these pointless pull requests to the repository. They serve no purpose other than to waste my time and your time.

Per

PR #4707, opened 2024-06-26
Update repositories.txt

sqmsmu:


per1234 (commented):
HI @sqmsmu. The change you propose here doesn't serve any meaningful purpose. We only accept changes to the contents of this repository when there is a valid reason for doing them so I will close this as invalid.

PR #4726, opened 2024-06-30
Update repositories.txt

maarten-pennings:


maarten-pennings:
Hmm, lib URL is not a clone URL

PR #4760, opened 2024-07-08
Update repositories.txt

HighASG936:


per1234 (commented):

@HighASG936 perhaps you actually intended to accomplish something productive with this pull request? If so, I am happy to provide advice on how you can accomplish whatever goal you might have. But I can only do that if you make the effort to clearly communicate your intent. You can do that by adding a detailed description in a reply here on the pull request thread. I'll continue to monitor the thread even though the PR is closed.

---

@HighASG936 I notice you have established a pattern of submitting invalid pull requests to this repository. Please start acting in a more respectful  manner in your activity in this repository.

We receive a huge amount of activity in this repository for thousands of libraries. Arduino makes the Library Manager distribution system available to the entire community rather than reserving it only for our own libraries. If every one of the 3rd party library authors behaved in as disruptive manner as you, we would be forced to no longer allow 3rd party library authors to use Library Manager because we simply don't have the resources to dedicate to cleaning up all that mess. So your behavior is actually very harmful to the entire community.

PR #4836, opened 2024-07-25
Update repositories.txt

Arathierl:


umbynos:
Hello @Arathierl, it seems that the library you are trying to submit is [causing some problems](https://github.com/arduino/library-registry/actions/runs/10089139998/job/27896099282#step:8:27) on our automated system:
the [library examples](https://github.com/Arathierl/Home-Automation/blob/main/example) must be placed in the **`examples/` folder**. Note that the examples folder name must be written exactly like that (with lower case letters). 
``` Home-Automation/examples/<example sketches> ```
I suggest to read the specifications that you can find here: https://arduino.github.io/arduino-cli/1.0/library-specification/

PR #4854, opened 2024-07-28
Move all davetcc repositories under TcMenu org

davetcc:
Moving all libraries into a GitHub organisation, same content as before (old links still work but better to have the registry pointing at the right place).


davetcc:
The rationale for this is that tcMenu has become rather large, and now has more than one dev/owner, we'd been planning the move to an organisation for a while, and while there may be a few old links (that still work) temporarily, long term everything will link to these organisation level repos.

Please do check this more carefully than most to ensure I have only updated my own libraries.

per1234 (commented):
Hello @davetcc. I am confirming your request.

umbynos (approved):

umbynos:
Solved in the linked PRs

PR #4896, opened 2024-07-31
change my url

vpbharath:


ERLtechtvm:
https://github.com/ERLtechtvm/RobotControl.git this is my changed url  can you please check

ERLtechtvm:
https://github.com/ERLtechtvm/RobotControl.git this is my changed url can you please check @ArduinoBot

repositories.txt
```diff
@@ -7218,5 +7218,4 @@ https://github.com/protheeuz/AsyncWebServerESP
 https://github.com/JoelJojoP/SimpleIMU
 https://github.com/vickash/RotaryEncoderPCNT
 https://github.com/kanitawa/RotEnc
-https://github.com/vpBharath/RobotControl.git
```
per1234, line 7221: You must not remove the URLs of other people's libraries. That is incredibly irresponsible and disrespectful behavior.

Please add the URL back.

per1234 (commented):
In addition to changing the URL, you also changed the library name.

The library name is defined by the `name` field in your library's `library.properties` metadata file:

https://github.com/ERLtechtvm/RobotControl/blob/main/library.properties#L1

```text
name=RobotControl
```

We do allow you to change the library name on request (though we discourage the practice as it makes work for us), but it is mandatory that every library have a unique name.

The new name you chose "RobotControl" has already been taken so you are not allowed to rename your library to this name.

Please change the value of the `name` field in your `library.properties` file to a unique name. You can search the [Arduino IDE Library Manager](https://docs.arduino.cc/software/ide-v2/tutorials/ide-v2-installing-a-library#installing-a-library) to check whether a name is already taken.

1. Change the value of the `name` field in your `library.properties` file to a unique name.
   **ⓘ** You can search the [Arduino IDE Library Manager](https://docs.arduino.cc/software/ide-v2/tutorials/ide-v2-installing-a-library#installing-a-library) to check whether a name is already taken.

Regards, Per

repositories.txt
ERLtechtvm: that is my old url  i change it with my new url do i detele it . this is my new url https://github.com/ERLtechtvm/RobotControl.git

per1234: Ah, I see now you have been changing your username many times. So I was wrong and it is OK to remove this URL.

However, you still must fix the problem of the non-unique library name.

PR #4901, opened 2024-07-31
Update repositories.txt

ErlTechnologies:
update the library

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -7221,3 +7221,4 @@ https://github.com/vickash/RotaryEncoderPCNT
 https://github.com/kanitawa/RotEnc
 https://github.com/vpBharath/RobotControl.git
 https://github.com/ERLtech/ERLtech-RobotControl.git
+https://github.com/ErlTechnologies/BTRobotControl.git


per1234 (commented):
Closing as duplicate of https://github.com/arduino/library-registry/pull/4900

@ErlTechnologies you are creating a **HUGE** amount of work for me with your unprofessional behavior. This is highly irresponsible.

We process many thousands of library submissions in this repository. Arduino has opened the Library Manager distribution service to the entire community instead of keeping it only for the distribution of our own libraries. If every one of the 3rd party library authors behaved the way you have been doing then we would be forced to stop offering this service to the community. So this type of behavior is very harmful to the Arduino community.

Although we strive to make the creation of Arduino projects accessible to absolutely everyone, creating and distributing a library to the community is something for experienced developers. So we expect a high level of competence, professionality, and respect from the people who use this repository. You have not shown anywhere near that level of competence, professionality, or respect in your behavior here. Do you really think it is a good idea for you to behave this way in a global community of professional software developers?

So you need to make a decision now. Either you are going to start behaving in a competent manner, or else you need to stop polluting this repository and the Arduino Library Manager with your incompetence. If this is the quality you bring to your libraries, then they won't be of any use to anyone, so what is the point of all this???

If your behavior doesn't improve immediately, you will be banned from all of Arduino's libraries and reported to GitHub.

PR #4911, opened 2024-08-02
Update repositories.txt

ErlTechnologies:
I added an new example code and update version


ErlTechnologies:
BTRobocontrol is my own library i didn't want to change it I added an new updation with this.

PR #4917, opened 2024-08-02
Add friends of audio-tools

egnor:
These are miscellaneous libraries from pschatzmann which work with his audio-tools library to supply various codecs and such.


egnor:
Removed the offending libraries @ArduinoBot

PR #4945, opened 2024-08-05
Update repositories.txt for new library

chiara-septentrio:
Added a new library for Septentrio GNSS


chiara-septentrio:
Updated the library with a tag @ArduinoBot

PR #4975, opened 2024-08-12
Remove EspNowNetwork

Johboh:
- `EspNowNetwork` used to be a "fat" library, which was built and published as individual libraries to ESP IDF Component Registry and Platform IO Registry using github actions. This, however, is not compatible with Arduino library registry, as each github repo need to be its own library. The `EspNowNetwork` repo has now been split up and replaced with four individual repositories instead:
  - https://github.com/Johboh/EspNowNetworkHost
  - https://github.com/Johboh/EspNowNetworkHostDriver
  - https://github.com/Johboh/EspNowNetworkNode
  - https://github.com/Johboh/EspNowNetworkShared

All of which are already registered in this registry.


per1234 (commented):
Hello @Johboh. I am confirming your request.

umbynos (approved):
Done in https://github.com/arduino/library-registry/pull/4976#pullrequestreview-2235204016

PR #4997, opened 2024-08-21
Update repositories.txt

jezhill:
added SyncGenie library


jezhill:
@ArduinoBot tag added

PR #5001, opened 2024-08-22
Update repositories.txt

Frallan08:


per1234 (commented):
Thanks for your interest in contributing to the Arduino Library Manager registry @Frallan08!

I see you are attempting to submit a library that does not meet the requirements for inclusion in the Arduino Library Manager registry:

Since a significant amount of work would need to be done to bring the library into compliance, I am going to close this.

Since this "library" is utterly trivial, I would recommend you to look through the libraries that are already in the Library Manager to see if one of them provides the equivalent functionality (perhaps even in a superior manner).

If you find that none of the existing libraries can provide this functionality, you are welcome to work to bring the library into compliance with all the requirements, and then submit another pull request here to make another try at adding the library to the registry.

PR #5025, opened 2024-08-24
R4A_I2C: Add Robots-For-All I2C support library

LeeLeahy2:


per1234 (commented):
Hello @LeeLeahy2. Thanks for your interest in contributing to the Arduino Library Manager index.

```text
A problem was found with your submission https://github.com/LeeLeahy2/R4A_I2C

:x: **ERROR:** The repository has no tags. You need to create a [release](https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) that matches the `version` value in the library's library.properties file.
```

Regards, Per

per1234 (commented):
@LeeLeahy2 I see that your repository doesn't even contain an Arduino library.

Arduino Library Manager is for use to distribute libraries that will be useful to the Arduino Library Manager. Clearly your library does not meet that standard since you haven't even started writing one.

For this reason, I will close this pull request as invalid.

I will also request that, in the future, you not abuse our Library Manager service. Rather than reserving it for exclusive use to distribute our own and partner libraries, Arduino has invested significant resources to open this service up to the entire community. In return, we expect the community to be respectful and responsible in the use of the service. You have not done that. We have thousands of library authors using this service. If all of them acted as you have done, we would be forced to close the registry. So this behavior is harmful to the entire community, and also doesn't benefit you an any way.

If you eventually create a library, and are confident that it is high quality, useful to the Arduino community, and well tested, then you are welcome to submit another pull request.

LeeLeahy2:
I have created the initial tag.  @ArduinoBot please review.

PR #5035, opened 2024-08-25
Update repositories.txt

Bhanuka-Anjana:
- Correct the mistake


per1234 (approved):
Thanks @Bhanuka-Anjana!

PR #5040, opened 2024-08-26
Update repositories.txt for AMY

bwhitman:


bwhitman:
@ArduinoBot added the release

PR #5042, opened 2024-08-27
R4A_Robot: Add Robots-For-All infrastructure and common support routines library

LeeLeahy2:


per1234 (commented):
Closing as duplicate of https://github.com/arduino/library-registry/pull/5025

@LeeLeahy2 did you read what I wrote there???

> @LeeLeahy2 I see that your repository doesn't even contain an Arduino library.
>
>Arduino Library Manager is for use to distribute libraries that will be useful to the Arduino Library Manager. Clearly your library does not meet that standard since you haven't even started writing one.
>
>For this reason, I will close this pull request as invalid.
>
>I will also request that, in the future, you not abuse our Library Manager service. Rather than reserving it for exclusive use to distribute our own and partner libraries, Arduino has invested significant resources to open this service up to the entire community. In return, we expect the community to be respectful and responsible in the use of the service. You have not done that. We have thousands of library authors using this service. If all of them acted as you have done, we would be forced to close the registry. So this behavior is harmful to the entire community, and also doesn't benefit you an any way.
>
>If you eventually create a library, and are confident that it is high quality, useful to the Arduino community, and well tested, then you are welcome to submit another pull request.

If you didn't read it, why not? You should be responsible enough to monitor the communications on your pull requests here.

If you did read it, then why did you attempt to submit the library again after I clearly explained that it was not acceptable?

Either way, you are behaving in a very disrespectful and irresponsible manner. If I see any more such behavior from you, I will have you banned from all of Arduino's repositories, remove all your libraries from the Library Manager, and report your account to GitHub. This is your only warning. It is time for you to either get your act together, or else stop interacting with Arduino's repositories.

PR #5047, opened 2024-09-01
Update repositories.txt

HackerNowful:


HackerNowful:
i just updated my Sd-Ducky to a new version like V1.0.1 its been already 2 months its not shown on arduino libraries
https://github.com/HackerNowful/SD-Ducky

HackerNowful:
@ArduinoBot

HackerNowful:
@ArduinoBot

per1234 (commented):
Hello @HackerNowful. When you submitted the library, the bot shared the URL of the dedicated logs page for the library:

https://github.com/arduino/library-registry/pull/4252#issuecomment-2014361691

> You can check the logs from the Library Manager indexer for your library(s) here:
http://downloads.arduino.cc/libraries/logs/github.com/HackerNowful/SD-Ducky/

The availability of the logs is also documented in the FAQ:

Did you bother to check the logs? If you did, you would see this:

```text
2024/09/01 10:35:20 Checking out tag: 1.0.1
2024/09/01 10:35:20 Release SD-Ducky:1.0.1 has wrong library name, should be sd-ducky-v1
```

The reason for this restriction is that this is the unique identifier for the library, meaning that any changes to the identifier might cause problems to users who rely on the established identifier in command line operations or for other library developers who have defined it as one of their dependencies.

Your library was registered under the name "sd-ducky-v1", but you later changed value of the `name` property in the `library.properties` metadata file from "sd-ducky-v1" to "SD-Ducky":

https://github.com/HackerNowful/SD-Ducky/blob/1.0.1/library.properties#L1

```text
name=SD-Ducky
```

This is the reason why the 1.0.1 release was rejected.

---

@HackerNowful you have [established a pattern of submitting pointless pull requests to this repository](https://github.com/arduino/library-registry/pulls?q=author%3AHackerNowful). Please start being more respectful in your behavior here. It seems you might be under the impression that this repository is maintained by an omniscient robot that can psychically determine what you are attempting to accomplish by submitting a PR. That is not the case here. Although we have automated the library submission process, all other maintenance of this repository is done by humans. Please act accordingly; as you would do if interacting with a human in person.

We are here to help you, but we can only do that if you make the effort to communicate in a way that a human can understand. If you think there is a problem with the library registry system, then you can submit an [issues](https://docs.github.com/issues/tracking-your-work-with-issues/about-issues) which clearly describes the problem in a way a human can understand.

If you can't behave respectfully in this repository then you will be banned from all of Arduino's repositories and your libraries will be removed from Library Manager.

PR #5090, opened 2024-09-11
Remove library for maintenance

IRDMPE:
Updating information


IRDMPE:
Removing library for re-registration. Please let me know if the correct way to do this. We needed to change the library names and clean up commit history. My understanding is that this requires re-registration for the changes to be correctly picked up.

IRDMPE:
If re-registration or registration updates can be done without removing, that would be great. The URL is still accurate, but the properties file, tagging, and commit history are edited.

per1234 (commented):
> We needed to change the library name

Please follow the standard procedure that is documented here:

PR #5122, opened 2024-09-20
Update repositories.txt

guerratron:
adding minIniFS


guerratron:
adding release, @ArduinoBot

PR #5154, opened 2024-09-24
added parameter and esp-now-midi libraries

thomasgeissl:
hey,
this add a parameter library and a midi over esp-now library.


thomasgeissl:
dear @ArduinoBot, you are right, i forgot to push my tags to the remote. scusi

PR #5252, opened 2024-10-16
Update repositories.txt

guerratron:
update TouchCal library


per1234 (commented):
@guerratron making pointless pull requests is _**NOT**_ the appropriate way to update your library. You need to take the time to read the documentation:

It is true that we have a crude automated system for handling submissions, but it isn't some magical psychic robot that is able to infer your wishes from a pointless PR and grant them like a genie. So spamming this repository with pointless pull requests only wastes my time and your time. Please don't do that.

PR #5264, opened 2024-10-21
Commit changes

lucweb:


per1234 (commented):
This proposal to move the URL from one arbitrary location in the file to another serves absolutely no purpose so we will not accept it.

@lucweb I have to assume that you hoped to accomplish something other than wasting my time and wasting your time by submitting this pull request. No matter what that thing you hoped to accomplish was, _**this is not the way to accomplish it**_. I'm happy to help you to accomplish that thing, but I can only help if you will make the effort to communicate to me as a human.

PR #5318, opened 2024-10-29
add https://github.com/bdlow/IO22_IO_Board

bdlow:
Drive the Eletechsup IO22 family of I/O boards: IO22D08 and IO22C04. arduino-linted for extra goodness.


bdlow:
added tag (well, I'd already added it, but hadn't pushed it... "By default, the git push command doesn’t transfer tags to remote servers")
@ArduinoBot

PR #5330, opened 2024-10-30
tar

gbitar1:
bz22


per1234 (commented):
Please don't submit pointless pull requests to this repository @gbitar1.

This is a free open source project that benefits a community of passionate creators. It is a lot of work to maintain and pull requests like this just consume limited resources that could be put towards important work. So it is really harmful when you do things like this.

Feel free to experiment all you like in your own repositories, but leave our repositories alone please.

PR #5355, opened 2024-11-05
Removed old pxserv library

pxsty0:


pxsty0:
Hi, we would like to remove the old libraries until the PxServ Arduino Libraries are published again in a new way, I would be glad if you return positively to our request.

PR #5357, opened 2024-11-05
Detele AvrFHT

ArthurKoba:
My open problem has not been solved since August, it is obvious that it is easier to remove the library and try to add it again.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -7372,7 +7372,6 @@ https://github.com/septentrio-gnss/Septentrio_Arduino_library
 https://github.com/iamfaraz/Waveshare_ST7262_LVGL
 https://github.com/Danzo-Systems/FM25060_Library
 https://github.com/DouglasFlores-fun/SimpleLogger
-https://github.com/KobaProduction/AvrFHT
 https://github.com/cgobat/spartan-edge-esp32-boot/
 https://github.com/sierramike/HomeAssistantMQTT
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Hello @ArthurKoba. I understand your frustration about our delay in accommodating your request (https://github.com/arduino/library-registry/issues/4977), but we must follow the standard procedure. We do not allow the removal of a library when the plan is to simply add it back again.

A strange problem was encountered when the database maintainer tried to perform the renaming (https://github.com/arduino/library-registry/pull/4978#pullrequestreview-2235229947). Unfortunately we have very limited resources and a vast number of tasks that need to be completed so we have not been able to perform the investigation into why this happened and how it can be fixed. Please be patient. I apologize for the inconvenience this has caused.

ArthurKoba:
Hello @per1234. Thanks for your answer. I understand that you have your own instructions and regulations, but my problem was not solved in 3 months. Why, if there are regulations and they are observed, are no actions taken to address them? And if they do, why are there no regulations for returning to unresolved problems? I really have a lot of questions.

I even wrote a month later and asked, but there was no answer, are there ways to speed up the process of solving my problem? If not, and there is no time, wouldn't it be simpler to actually remove the library and add it again? This will at least give some result than meaningless compliance with regulations. Understand that because of this problem, not only I am experiencing difficulties, but also people who use this software at my prompting.

PR #5411, opened 2024-11-13
Removing deprecated library

maker-zone:
The same functionality of sharing signatures between multiple libraries to make them interchangeable can be achieved without using an interface, so this library was deprecated to avoid wasting space on an arduino.


per1234 (commented):
Hi @maker-zone . I am confirming your request.

Unlike submissions, this type of request is still handled manually. The Library Manager index maintainer occasionally comes through to clear all the outstanding requests. They will close this issue when it is done. You will get a notification from GitHub at that time.

Information about that here:

maker-zone:
I removed one library
UltiBlox DisplayValueCommon
...and just now added another
UltiBlox-SerialCommands
It looks like the bot is thinking I just renamed it.
What's the best way to resolve this? Should I just wait until the previous removal is completed?

per1234 (commented):
@maker-zone please submit a separate pull request for each distinct request.

You can have multiple open pull requests by creating a separate [branch](https://docs.github.com/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches) in your [fork](https://docs.github.com/get-started/quickstart/fork-a-repo) for the specific changes you want to propose, then submitting a pull request from that branch.

In order to facilitate this workflow, you should never use the `main` branch of your fork to stage changes for a pull request. Instead, the `main` branch should be kept in sync with the parent repository's branch, so it can be used as the base for your pull request branches.

Please remove the change you made to this pull request so that it only contains the change of removing `https://github.com/UltiBlox/DisplayValueCommon`.

maker-zone:
Ah I think I understand my mistake now.
I've removed that addition from this. I hadn't intended to include it in the same PR but didn't realise that it would get included automatically.
From now on I'll use branches other than main for each change.

MatteoPologruto (approved):
Done:
```
2024/11/26 15:16:39 Loaded 7503 libraries from DB
Removing UltiBlox DisplayValueCommon
Success!
```

PR #5368, opened 2024-11-06
Update repositories.txt

pxsty0:

PR #5416, opened 2024-11-13
Removed PxServ Arduino Library

pxsty0:


per1234 (commented):
Hello @pxsty0. Please add a comment that provides a detailed explanation of why you want the library to be removed.

We are happy to accomodate removal requests in cases where the library maintainer wants to permanently remove a library from Library Manager. However, I get the feeling you are up to some monkey business here because we just got done updating the library's URL (https://github.com/arduino/library-registry/pull/5368), so it doesn't make sense that you would care enough about the library's registration to want to keep its URL updated, but also want it removed.

In these cases, we often find that the library maintainer doesn't actually plan to permanently remove the library, but is instead planning to submit the library again after it is removed as a hacky attempt at overcoming some problem they are having. This is not allowed (and most often wouldn't be effective or efficient anyway). We only allow removals if they are to be done on a permanent basis. I am happy to advise you on the proper way to accomplish other goals, but I can only do that if I understand what you are trying to accomplish.

pxsty0:
I would like to remove the library as the information of the old library is still visible in the library manager. Some time after the library is completely removed, I will republish it as a library that has never been published before.

pxsty0:
what I actually want is to remove all previous versions of the library from the library managers and then release it as a new library

pxsty0:
I sorry for the inconvenience; however, I had to use this method as no change was visible when I checked (https://downloads.arduino.cc/libraries/logs/github.com/pxsty0/pxserv.arduino). @per1234

pxsty0:
@ArduinoBot

pxsty0:
The updates have started to appear in the library managers. Thank you for your support.

PR #5425, opened 2024-11-15
Update repositories.txt to add lautisilber/SmartComm library

lautisilber:
Thank you dearly for all your work maintaining the Arduino project!


lautisilber:
What about now? (Sorry for the inconvenience btw)

PR #5493, opened 2024-11-29
EmotiBit KTD2026

nitin710:
- Add EmotiBit KTD2026 to the library-registry


nitin710:
Created a release. @ArduinoBot

PR #5496, opened 2024-11-30
v1.0.1

RASPIAUDIO:
@ArduinoBot


RASPIAUDIO:
@ArduinoBot

per1234 (commented):
@RASPIAUDIO I'm guessing you are attempting to publish a new release of your library, which has already been submitted to the Arduino Library Manager Registry.

This means there is no need to take any additional action to add new releases of your library to Library Manager. They will be added automagically! So please don't submit pull requests when you release a new version of your library. These pull requests are pointless and only waste your time and my time.

PR #5578, opened 2024-12-11
Update URL "NH8CHIR-lib"

Karibura-Cyber:


per1234 (commented):
@Karibura-Cyber you have established a pattern of submitting this type of pointless pull request. Please don't do any more. It only wastes your time and wastes my time.

I have to assume you are trying to accomplish something productive with all these pull requests. I'm here to help you, but I can only do that if you will make the effort to communicate what it is you are trying to accomplish. If you will add a comment here on the pull request thread telling me what you hope to achieve, I'll be happy to provide assistance.

PR #5624, opened 2024-12-19
Update repositories.txt

digitalfen:


per1234 (commented):
Hi @digitalfen. I see this pull request removes your library URL from the list. This seems very strange to me because you added the URL only minutes ago (https://github.com/arduino/library-registry/pull/5622).

This makes me suspect that you are trying to accomplish something other than removing the library. I'm here to help, but I can only do that if you will make the effort to communicate what you are doing to me as a human.

So please tell me what was the reason for this pull request so that I can be sure to handle the request correctly.

I'll also request that you avoid submitting duplicate pull requests. The automated system is set up to allow us to accomplish tasks with a single pull request, and the bot explains this in it comments, but you still submitted multiple pull requests. Doing that makes work for me and also is inefficient for you.

PR #5637, opened 2024-12-21
Update repositories.txt

dejwk:


dejwk:
@ArduinoBot, fix applied.

PR #5699, opened 2024-12-31
Update repositories.txt

ajangrahmat:
Delete My Repository

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -7656,7 +7656,6 @@ https://github.com/moononournation/Dev_Device_Pins.git
 https://github.com/ELOWRO/ADS1119
 https://github.com/kelasrobot/KelasRobotIO
 https://github.com/rescenic/rescenicio
-https://github.com/kelasrobot/FonnteArduino
 https://github.com/JokoArdh/inIo
 https://github.com/robbywm/RobbyIO/
 https://github.com/cakraawijaya/MQ2_LPG


per1234 (commented):
@ajangrahmat please provide a detailed description of why you are requesting the removal of the library.

You submitted the library only 17 hours ago (https://github.com/arduino/library-registry/pull/5682), then you submitted a url change request (https://github.com/arduino/library-registry/pull/5685), and now you are already requesting the removal. This makes me think that you don't actually want to remove the library, but instead think you can accomplish something by removing the library and then immediately adding it back again. We only allow removals when the intent is a permanent removal. We don't allow it for some hacky attempt at trying to work around some perceived limitation in the system.

per1234 (commented):
@ajangrahmat hasn't been responsible enough to respond, but they have been busy submitting even more time wasting pull requests for this library. This makes it clear that the suspicions I voiced in my previous review were correct and @ajangrahmat did not have any intention to permanently remove the library, but was instead only wasting my time.

@ajangrahmat you are establishing a pattern of unprofessional behavior in this repository. It is time for you to start to behave in a respectful manner. If I see any more misbehavior from you, I will remove all your libraries from Arduino Library Manager and permanently ban you from this repository. Do you really think it is a good idea to behave this way when you are trying to run a technology business? GET YOUR ACT TOGETHER!!!!

PR #5736, opened 2025-01-09
Update repositories.txt

YoavPaz:
@ArduinoBot


YoavPaz:
@ArduinoBot

YoavPaz:
@ArduinoBot

YoavPaz:
@ArduinoBot

per1234 (commented):

@YoavPaz  please don't submit pointless pull requests to this repository. The Arduino Library Manager Registry contains thousands of libraries from hundreds of authors. Finite resources are available to maintain this resource which is so important to the Arduino community. Pointless pull requests waste those resources and so this is irresponsible behavior. Repeated irresponsible activity results in the revocation of Library Manager Registry privileges.

PR #5778, opened 2025-01-15
Update repositories.txt

cakraawijaya:
I want to delete my library topic


per1234 (commented):
Duplicate of https://github.com/arduino/library-registry/pull/5781

@cakraawijaya in the future please only create one pull request for each operation.

As was already clearly explained in the comments from the bot, the automated submission system is designed to allow you to trigger additional rounds of checks on the library in the original pull request after fixing problems reported by the bot. So there is no need to create additional pull requests and doing so only makes work for the human maintainer of this busy repository.

---

⚠ Repeated irresponsible activity may result in the revocation of your Library Manager Registry privileges.

---

PR #5788, opened 2025-01-17
Update repositories.txt

codewitch-honey-crisis:
added several libs

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -7693,5 +7693,5 @@ https://github.com/codewitch-honey-crisis/htcw_bits
 https://github.com/codewitch-honey-crisis/htcw_data
 https://github.com/codewitch-honey-crisis/htcw_io
 https://github.com/codewitch-honey-crisis/htcw_ml
-https://github.com/codewitch-honey-crisis/htcw_gfx
-https://github.com/codewitch-honey-crisis/htcw_uix
+https://github.com/codewitch-honey-crisis/gfx
+https://github.com/codewitch-honey-crisis/uix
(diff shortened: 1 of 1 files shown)


codewitch-honey-crisis:
It says my libraries are missing library.properties but they aren't?

codewitch-honey-crisis:
fixed some urls. still have the problem with it not finding my library.properties files.

codewitch-honey-crisis:
These are relatively minor warnings. Some of these are basically private libraries I made to avoid duplicating code between major libraries which is why they don't have examples. What exactly are the expectations here? Am I just being impatient or this actually blocked on those warnings? Thanks.

per1234 (commented):
Hello @codewitch-honey-crisis. I see you have fixed the error reported by the bot. However, Library Manager works exclusively from Git tags, so the bot won't be able to see the fix you made until you create a new [release](https://docs.github.com/en/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the library repository.

https://github.com/codewitch-honey-crisis/htcw_data/tree/1.4.2

https://github.com/codewitch-honey-crisis/htcw_ml/tree/0.2.3

https://github.com/codewitch-honey-crisis/htcw_io/tree/1.2.2

(note that the `library.properties` metadata was not present in the repository at the tagged point in the revision history)

Regards, Per

codewitch-honey-crisis:
ah okay. I meant to tag those anyway. Thanks. And sorry about earlier. I was super frustrated by the time I filed that bug report and I needed to not be near a computer at that point.

codewitch-honey-crisis:
@ArduinoBot I've updated my tags.

PR #5834, opened 2025-01-27
update links to some mathieucarbou repos

mcendu:
cc @mathieucarbou

Two repositories maintained by mathieucarbou are [moved to the organization `ESP32Async`](https://github.com/mathieucarbou/ESPAsyncWebServer/blob/main/README.md), but the Library Registry is not updated. This commit updates the registry.


mcendu:
Overlaps with scope of #5907 & #5808. Closing.

PR #5845, opened 2025-01-29
Simplify registry validation check workflows

cmaglie:
This PR changes the GitHub Action workflow that makes the validation check of the registry file.

Previously a golang project made for this purpose was built and run. Now it uses directly the library repository engine to perform this task. This change will simplify the maintenance of this repository since we no longer need to maintain a separate golang project.


alessio-perugini (approved):
LGTM

per1234 (changes requested):
These are now obsolete/superfluous and so should be removed:

- https://github.com/arduino/library-registry/blob/use-new-lib-engine/.github/workflows/check-python-task.yml
- https://github.com/arduino/library-registry/blob/d89b0e9211d952d85dbb8f28e4b98214b9a01af5/Taskfile.yml#L60-L75
- https://github.com/arduino/library-registry/blob/d89b0e9211d952d85dbb8f28e4b98214b9a01af5/README.md?plain=1#L6
- https://github.com/arduino/library-registry/blob/d89b0e9211d952d85dbb8f28e4b98214b9a01af5/README.md?plain=1#L13
- https://github.com/arduino/library-registry/blob/d89b0e9211d952d85dbb8f28e4b98214b9a01af5/pyproject.toml#L15-L19
  (and run `poetry lock` to sync the lockfile)
- https://github.com/arduino/library-registry/blob/use-new-lib-engine/.gitignore (at least the obsolete entries)

per1234 (approved):
Thanks Cristian!

PR #5854, opened 2025-01-31
Update repositories.txt

pmarques-dev:


pmarques-dev:
sorry, tag added, @ArduinoBot

PR #5883, opened 2025-02-06
Update repositories.txt

kotwatthana:
add TimerOne_V2


per1234 (commented):
@kotwatthana I already provided assistance in the first invalid request you made about this library:

https://github.com/arduino/library-registry/issues/5870#issuecomment-2635281803

Did you read what I wrote there? If so, then why are you still submitting useless requests? I already offered to provide assistance if you have any problems or questions while following those instructions. I'm here to help, but I can only do that if you will make the effort to communicate to me as a human being!

Please don't submit pointless pull requests to this repository. The Arduino Library Manager Registry contains thousands of libraries from hundreds of authors. Finite resources are available to maintain this resource which is so important to the Arduino community. Pointless pull requests waste those resources and so this is irresponsible behavior.


---

---

PR #5890, opened 2025-02-07
Update library

nextgentech-512:
Library Update


helloworld179 (approved):

per1234 (commented):
@nextgentech-512 if you want to update the library then you must follow the instructions we provide in the documentation:

Submitting this sort of pointless pull request doesn't accomplish anything other than wasting your time and wasting my time.

Please don't submit pointless pull requests to this repository. The Arduino Library Manager Registry contains thousands of libraries from hundreds of authors. Finite resources are available to maintain this resource which is so important to the Arduino community. Pointless pull requests waste those resources and so this is irresponsible behavior.

---

If you are having some difficulty with updating your library, I am happy to provide assistance. But I can only do that if I understand the situation. You can add a detailed description in a reply here on the pull request thread. I'll continue to monitor the thread even though the PR is closed.

PR #5906, opened 2025-02-11
Update repositories.txt

roncoa:
https://github.com/roncoa/StringEEPROM

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -1,8 +1,8 @@
+https://github.com/roncoa/StringEEPROM
 https://github.com/VanSilver/ShiftRegisterPWM595
 https://github.com/AryanBhirud/BitFlash_Client
 https://github.com/a-saab/PrayerTimes
 https://github.com/nextgentech-512/Single-Wire-Data-Bus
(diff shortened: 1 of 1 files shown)


per1234 (commented):

@roncoa please don't submit pointless pull requests to this repository. The Arduino Library Manager Registry contains thousands of libraries from hundreds of authors. Finite resources are available to maintain this resource which is so important to the Arduino community. Pointless pull requests waste those resources and so this is irresponsible behavior.

---

I guess you hoped that this would somehow magically fix the problem you are having with your library release. Pointless pull requests will never accomplish anything other than wasting your time and wasting my time.

We are happy to provide assistance. You only need to make the effort to communicate, as you did in https://github.com/arduino/library-registry/issues/5907.

If you think there is a problem with the Library Manager registry system, please do submit an issue.

If you are having difficulty with a pull request, you can add a comment on the PR comment thread with a detailed explanation of the problem. I monitor this repository so I will see your comment and respond.

For general discussion or project assistance, you can post on Arduino Forum:

I also monitor the forum and there are a lot of other knowledgeable forum members too.

PR #5925, opened 2025-02-14
Update repositories.txt

braydenanderson2014:
Add an Optional implementation to the Arduino Library Manager!


braydenanderson2014:
@ArduinoBot

PR #5943, opened 2025-02-16
Add ArduinoMessageBus Library

yunussandikci:
Hello Team,  
I would like to propose adding my open-source library to the registry. It is straightforward but I believe it will simplify the process of gathering multiple values from various sources, such as sensors.  
Thank you!


yunussandikci:
@ArduinoBot Please check

PR #5944, opened 2025-02-16
Removed Grafik Serial Lib

adityasharma-tech:


per1234 (commented):
Hi @adityasharma-tech. Please provide a detailed explanation for why you are requesting the removal of the library.

We do accommodate removal requests when the library maintainer wishes to permanently remove the library from Arduino Library Manager. However, I see you submitted this library only days ago (https://github.com/arduino/library-registry/pull/5917). It would be quite strange to immediately decide you want to permanently remove the library immediately after submitting it.

We sometimes see people try to remove and then resubmit libraries as a hacky way to accomplish something. Doing that is not allowed, and likely would not accomplish anything other than wasting everyone's time anyway.

PR #5945, opened 2025-02-16
Add new library 'SmartSystem' to the library registry - simplified RPC over Wire and RF24

mgfryan:
Add new library 'SmartSystem'. Readme below:

# SmartSystem
*SmartSystem* is an Arduino library which simplifies building modular networks of Arduinos.

Functionality can be exposed by a Arduino module, and discovered, accessed, and executed remotely over either Wire or RF24 radio.

This library evolved from use of Arduinos in escape room setups, which required many Arduinos to be centrally controlled and monitored.

## Installation
Use the Arduino library manager to install *SmartSystem* library.

**At least 2** Arduinos are required to create a *SmartSystem* network.

### If using Wire protocol
I2C Wire connections **should use the default SDA and SCL pins for your Arduino** (e.g. A4 / A5 for a Nano). This is not something that can be configured when using this library.

Using the ```Wire_Settings``` parameter at startup, it is possible to override the Wire connection speed.

### If using an RF24 radio device (such as NRF24L01)

The default settings are those for UNO / Nano:

| Connection | Pin |
|------------|-----|
|CE|9*|
|CSN|10*|
|SCK|(connect to board default - 13 for Nano)|
|MISO|(connect to board default - 12 for Nano)|
|MOSI|(connect to board default - 11 for Nano)|
|INTERRUPT|2*|

Using the ```Radio_Settings``` parameter at startup, it is possible to override the pins marked with * in the table above. **Note that using an INTERRUPT pin is absolutely required by this library!**

```Radio_Settings``` also let you control other selected parameters of the RF24 device, such as ```RADIO_SPEED```, ```RADIO_POWER``` and ```CHANNEL``` (which defaults to 108), 

In general, it is best to stick to the defaults where possible. This library generally favours simplicity and consistency over configurabilty!

## Usage

The system setup always involves:

* The *SmartModules* - These are Arduinos acting as providers of functionality for the network. There can be up to 64 of these on a single network.
* A *SmartController* - A single main Arduino acting as system orchestrator

A *SmartModule* must always have **strictly one** communication method with the network - either Wire or RF24 Radio. This is selected at module startup using either ```startOverWire()``` or ```startOverRadio()``` function.

The *SmartController* can use either communication method, or both at once. It can see and control all connected modules, regardless of the communication protocol they are using. This is selected at controller startup using the ```usingWire()``` and ```usingRadio()``` methods.

#### Simple SmartModule Arduino sketch, which provides a function that returns an integer, exposed over RF24 radio
```C++
#include <SmartSystem.h>

// the number we want to return
volatile int answer = 42;

// Function we want to expose from the Arduino.
int theAnswer()
{
  return answer; // returns the answer variable
}

// sketch setup - set up the smart module here
void setup()
{
  // set up our Arduino as a smart module
  SmartModule
    .expose(theAnswer, "GetTheAnswer") // expose the function 'theAnswer', visible as 'GetTheAnswer'
    .addTag("Everything")              // tag the module (optional!) 
    .startOverRadio(1, "DeepThought"); // start as module 1 over Radio, with name "DeepThought"
}

// main sketch loop - nothing required here
// but code in loop() could, for example, read from a sensor, 
// and modify the variable to change the response
void loop()
{
}
```

#### Simple SmartController Arduino sketch, which accesses the SmartModule's functionality over radio
```C++
#include <SmartSystem.h>

void setup() 
{
  // start up serial
  Serial.begin(9600);
  while (!Serial);

  // start up the controller, on both radio and wire, and scan for modules
  SmartController.usingRadio().usingWire().scan();

  // Find the desired module and function on the network
  // and execute it, expecting an integer return value
  int result = SmartController
                .module("DeepThought")
                .function("GetTheAnswer")
                .runToGet<int>();
  
  // print the output
  Serial.print("Invoke by name: GetTheAnswer returned: ");
  Serial.println(result);
}

void loop()
{  
}
```

### Logging
To see more info on startup / connection issues, ```SmartController.doSerialErrorLogging(true)``` can be called at startup, which activates serial logging of any errors.

### Gotchas
Because the exposed functions on the modules are invoked from interrupt callbacks, the following rules apply:

* Do not use Serial logging within the exposed functions
* Do not carry out any long-running operations in the exposed functions - delegate any work to the main loop.
* Pass information between exposed functions and the rest of the sketch using variables marked as ```volatile```

If you fail to observe the above, things can fail in unexpected and unpredictable ways.

### Further information
The provided [examples](examples) cover most use cases.

The [AllMethodTypesModule.ino](examples/AllMethodTypesModule/AllMethodTypesModule.ino) sketch shows how all different function types can be exposed from a module.

The [UniversalController.ino](examples/UniversalController/UniversalController.ino) is an interactive controller sketch, which shows how to scan, iterate over modules and functions, extract metadata from them, and invoke methods with all the allowed function signatures.

Using the sketches above on 2 Arduinos demonstrates all the possible ways of calling functions using a *SmartSystem*, and constitutes a good test when setting up a new system.

## Contributing

Pull requests are welcome!

## License

[MIT](LICENSE.txt)

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -7801,3 +7801,4 @@ https://github.com/mrrwa/SFR_Reader
 https://github.com/ArtronShop/ArtronShop_LineMessaging
 https://github.com/boerge1/rfid
 https://github.com/yunussandikci/ArduinoMessageBus
+https://bitbucket.org/mgf_ryan/smartsystem


mgfryan:
@ArduinoBot: Added tag!

PR #5993, opened 2025-02-26
Adding a new one, removing another

tabahi:
- Added a new StatefulGSMLib for SIM800 modules.
- TabahiConsole needs to be removed since its hosting service is no longer operational.


per1234 (commented):
Hi @tabahi. Please use a separate pull request for each distinct operation:

- One pull request for the removal of the "TabahiConsole" library
- One pull request for the submission of the "StatefulGSMLib" library

The reason this is important is because we have an automated system for handling submissions, but removals must be handled manually. If you unnecessarily jam both operations into  single pull request then it means the automated system can't handle your submission.

PR #5997, opened 2025-02-27
Remove TabahiConsole

tabahi:
Not maintained. Dead hosting service.


MatteoPologruto (approved):
Done:
```
2025/03/12 09:32:33 Loaded 7831 libraries from DB
Removing TabahiConsole
Success!
```

PR #6006, opened 2025-02-28
Update repositories.txt

RobTillaart:
Add Arduino library for ADG2188 8x8 (cross point) matrix switch with I2C.


RobTillaart:
@ArduinoBot

PR #6043, opened 2025-03-14
Update repositories.txt

Arturo-JS:


per1234 (commented):
For the sake of a tidy repository, I'm going to close this due to lack of action to resolve the problem reported by the bot.

@Arturo-JS if you truly do want to submit this library to Library Manager, you are welcome to submit another pull request, this time making sure to follow the procedure correctly, and to be prepared to follow up on any problems that are found.

The invalid nature of this pull request, and your other pull request https://github.com/arduino/library-registry/pull/6044 makes me suspect that you submitted these pull requests as some sort of a hacky way to accomplish an unspecified personal goal (i.e., "I don't know how to install these libraries for my project, so I guess I need to submit them to the Library Manager registry"), rather than because you feel they should be included in Library Manager for the benefit of the entire Arduino community (which is the valid reason for submitting libraries for registration here). If I am right, then you should instead request assistance over on Arduino Forum:

https://forum.arduino.cc/

That will be the effective way for you to progress with your project, while submitting invalid pull requests here will only be a waste of time.

I'll continue to monitor this thread, so feel free to let me know if you have any questions or concerns.

PR #6048, opened 2025-03-14
Add SensEdu-Library

vladysor:
I would like to add SensEdu-Library, which is framework for Arduino GIGA R1 WiFi shield called SensEdu, allowing to use Arduino for radar and communication systems.


vladysor:
@ArduinoBot fixed paragraph

PR #6053, opened 2025-03-15
Add PIDEasy PID library to Arduino Library Manager

ByteVoltaTeam:
This is a simple PID control library with support for Proportional, Integral, Derivative control, windup constraint, and derivative smoothing.


ByteVoltaTeam:
@per1234 can you help me to solve the conflict, because i don't understand what is wrong

PR #6065, opened 2025-03-16
Update library PIDEasy

ByteVoltaTeam:
New version of library


per1234 (commented):
Hello @ByteVoltaTeam . After the initial addition of the library, the indexer job periodically checks each of the libraries in the Arduino Library Manager Registry and picks up any new release/tag that meets the requirements listed here:

---

Submitting a pull request when you want to publish a new release of your library will not accomplish anything other than wasting your time and wasting the time of the human maintainers of this repository. So please don't submit pointless pull requests here.

PR #6001, opened 2025-02-27
Renamed https://github.com/dac1e/Dcf77Receiver to https://github.com/…

dac1e:
…dac1e/DCF77RX

Renamed the repository, because it shall appear along with other repositories that all start with upper case letters 'DCF' instead of 'Dcf'. Along with this change, the appendix 'Receiver' was shortened to 'RX'.

PR #6124, opened 2025-03-28
Deleted DCF77RX

dac1e:
Some weeks ago I renamed this library from Dcf77Receiver to DCF77RX. After my pull request was merged and some weeks of waiting, the old library still appears in Arduino IDE, and the new renamed one doesn't. Hence I am removing this library.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -7875,7 +7875,6 @@ https://github.com/4211421036/gsr-arduino
 https://github.com/4211421036/IoTModule
 https://github.com/Elrindel/SomfyReceiver
 https://github.com/thebestia90/AD5259
-https://github.com/dac1e/DCF77RX
 https://github.com/ImSpeddy/L298Nlib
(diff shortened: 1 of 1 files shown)


per1234 (commented):
>Some weeks ago I renamed this library from Dcf77Receiver to DCF77RX. 

This is false. What you did was update the repository URL: https://github.com/arduino/library-registry/pull/6001

The library name is defined by the `name` properties of the library's `library.properties` file:

https://github.com/dac1e/DCF77RX/blob/main/library.properties#L1

The repository URL has no effect on the library name, so updating the URL is not sufficient to rename your library.

If you want to rename your library, you must follow the standard procedure for doing that, which is described in the documentation:

Please follow the correct procedure. Library removals are only allowed when you intent to permanently remove the library from the repository. We do not allow library removals as a hacky way to accomplish some other maintenance task. Since this pull request is not intended to permanently remove the library, I am closing this as invalid.

PR #6192, opened 2025-04-14
Update repositories.txt

Eason-SYC:
Supports displaying multiple colors and allows using colored lights to draw simple graphics.


Eason-SYC:
@ArduinoBot ok

PR #6241, opened 2025-04-26
Update repositories.txt

eyr1n:
Added https://github.com/eyr1n/esp32-ps3

The original repository (https://github.com/jvpernis/esp32-ps3) may no longer be maintained, so I created a forked library.
The fork fixed a problem with building with the latest ESP32 board library (v3.2.0).


eyr1n:
@ArduinoBot

PR #6244, opened 2025-04-26
feat: add a light Task scheduler

belem2050:
A minimal Task scheduler for beginners.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -7996,3 +7996,4 @@ https://github.com/Arduino15/A15RGB
 https://github.com/eyr1n/esp32-ps3
 
 https://github.com/belem2050/TaskScheduler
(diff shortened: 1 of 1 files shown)


belem2050:
@ArduinoBot

per1234 (commented):
Closing as duplicate of https://github.com/arduino/library-registry/pull/6242.

I see that you renamed the repository from `TaskScheduler` to `LightTaskScheduler` since that time. The existing registration will still work because GitHub automatically sets up a redirect from the old URL to the new URL when you rename a repository.

It is a good idea to avoid a reliance on such redirects in case they might stop working at some point in the future. So you are welcome to submit a URL update request. That should be done by following the procedure described here:

PR #6277, opened 2025-05-06
Update repositories.txt

ahagelberg:
Added my new PropulsionStepper library that allows both absolute and continuous rotation of stepper motors with acceleration. No other library allows for both options with acceleration.


ahagelberg:
@ArduinoBot Updated the category and license fields.

PR #6285, opened 2025-05-09
Update repositories.txt

codewitch-honey-crisis:
added htcw_mpm_parser


codewitch-honey-crisis:
@arduinoBot am i just being impatient? I've never had to have this reviewed by a human before i think?

PR #6403, opened 2025-06-07
Update repositories.txt

manoj26may:


per1234 (commented):
Hello @manoj26may. Please add a comment here on the pull request thread that provides a detailed explanation for why you wish to remove this library.

We do remove libraries on request, but I find it strange that you would request the removal after you submitted it only yesterday (https://github.com/arduino/library-registry/pull/6396). This makes me think that your true goal is not actually the permanent removal of the library, but instead that you plan to remove and then re-add the library in hopes that doing so will accomplish some other goal. That is not allowed and would not be productive anyway. We only allow removal requests when the intent is the permanent removal of the library.

per1234 (commented):
For the sake of a tidy repository, I'm going to close this due to the failure to provide the information I requested.

@manoj26may I'll continue to monitor this thread, so if you decide you want to provide the explanation I requested, then just comment here and I'll see it.

In the future, please take care to behave responsibly in your activity in this repository. The resources available for maintaining this project are limited. The Arduino Library Manager system is a very important resource for the entire Arduino community, so activity such as submitting pointless pull requests, and then ignoring requests for information you should have provided to begin with is very harmful.

PR #6417, opened 2025-06-10
WIP: Added SegLCDLib for segment LCD displays

petrkr:
Does not met release/tag yet as I still have it in development, but it can be noticed by Arduino users, so they can help with first real release.

For now I just want claim this name because it should be universal one and do not want to claim controller name as HT1621 developer did and that library you can not use with any other LCD than he used from Aliexpress, because library itself is actually not for HT1621, but for some random LCD, which by random uses this chip.

So I decided create more universal library which can be easily extended with other controllers and LCD displays.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -8106,3 +8106,4 @@ https://github.com/manoj26may/IOT
 https://github.com/zukardex/DigiPIN
 https://github.com/avisha95/AViShaOTA
 https://github.com/GabyGold67/ShiftRegGPIOXpander_AVR
+https://github.com/petrkr/SegLCDLib


per1234 (commented):
Hello @petrkr. Thanks for your interest in contributing to the Arduino Library Manager.

Please only submit libraries here after they are ready do be added to Library Manager. We receive thousands of pull requests in this repository, so we can't allow an accumulation of pull requests that can't be merged.

So I'm going to close this, but please do submit another pull request once the library is ready.

As for your idea of recruiting contributors from the Arduino community, the appropriate place to do that is Arduino Forum:

PR #6429, opened 2025-06-13
Update repositories.txt

RomanZ1243:


per1234 (commented):
Closing as duplicate of https://github.com/arduino/library-registry/pull/6430

@RomanZ1243 in the future please only open a single pull request for each submission. As I already explained in my previous comment, the submission handling system is designed to allow you to use a single pull request to work with the bot to get the library to the state where it is passing checks. So it is not necessary to open a new pull request in the case where your library initially fails the checks, and doing so only wastes your time and my time.

PR #6491, opened 2025-06-24
Add DMD32Plus to Repository List

ahmadfathan:
I have enhance a library to control dot matrix display with ESP32. It does not support ESP32 with latest version (3.x.x). The library is no more maintained. So I made new one with bug fixes and new feauture.


ahmadfathan:
Updated @ArduinoBot

PR #6527, opened 2025-07-02
A4988 stepper motor driver control library

k-off:
Add A4988 stepper motor driver control library

 - basic control
 - extended control
 - multiple steppers synchronization
 - non-blocking implementation


k-off:
Changes repo visibility @ArduinoBot

PR #6623, opened 2025-07-18
Update repositories.txt

John-Karatka:
Added MAX6953 dot matrix display driver library


John-Karatka:
Added tag @ArduinoBot

PR #6630, opened 2025-07-20
Update repositories.txt

max22-:
Added ble-keyboard-mouse-client library


max22-:
hi @ArduinoBot, can you re-check my submission ?

PR #6651, opened 2025-07-25
Add LX16A-bus Library

alecxcode:
A lightweight object-oriented library with a class for controlling serial bus servo motors (Hiwonder, LewanSoul) using a BusLinker. Designed to work with LX-16A Bus Servo.

**Features**:
- Compatible with 3-pin LX-16A bus servo motors, and possibly with other servos of this family
- Provides an easy-to-use class with methods
- Contains [examples](https://github.com/alecxcode/LX16A-bus/blob/main/examples) with smooth motion (easing in and out) in servo and motor modes

Before release of this library, there was no library compatible with the BusLinker (also known as the TTL/USB Debug Board for Serial Bus Servo) in the Arduino Library Manager. The BusLinker is a popular and easy to use board for LX-16A servos.  
The official Hiwonder/LewanSoul library is compatible with BusLinker, but it lacks object-oriented design, harder to use, unclear how licensed, and was never integrated into the Arduino Library Manager.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -1,3 +1,4 @@
+https://github.com/alecxcode/LX16A-bus
 https://github.com/akarimiz1/E62Module
 https://github.com/NartemAI/NaiPot
 https://github.com/exosite/arduino_exosite_http


alecxcode:
@ArduinoBot, a tag was added

PR #6657, opened 2025-07-27
Update repositories.txt

dayatkun01:


dayatkun01:
@ArduinoBot

dayatkun01:
> Hello @dayatkun01. I see you have fixed the error reported by the bot. However, Library Manager works exclusively from Git tags, so the bot won't be able to see the fix you made until you create a new [release](https://docs.github.com/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the library repository.
> 
> If you look at the repository contents at the latest tag, you can see that the version the bot is checking still has the problem:
> 
> https://github.com/dayatkun01/ClearDS1302/tree/arduino
> 
> (note that the repository did not contain any library code files at that tagged point in the revision history)
> 
> ❗ Remember to update the value of the `version` property in the library's [`library.properties` file](https://arduino.github.io/arduino-cli/latest/library-specification/#libraryproperties-file-format) as needed before making a release/tag.
> 
> Once you have done that, simply mention **@ArduinoBot** in a comment here to trigger it to run the automated checks again. After those are passing, your library will be automatically added to Library Manager.
> 
> If you have any questions or concerns, please let me know.
> 
> Regards, Per

Thank you so much for you helpful reply, i very helped because of you reply! i did not know if my not accepted if you did not tagged me. thank you so much!

PR #6689, opened 2025-07-31
Update repositories.txt

Nezaemmy:


Nezaemmy:
I don't see where the error came from; please help me to locate exactly the issue. Thank you

per1234 (commented):
Hello @Nezaemmy. Thanks for your interest in contributing to the Arduino Library Manager.

The cause of the error reported by the bot is that your library repository does not meet this requirement:

> - [ ]  The library.properties file must be located in the root of the repository.

If you look at the root of your repository

https://github.com/Nezaemmy/NezaButton/

, you will see that the problem is the library is not directly under the root of the repository, but is instead in a subfolder:

https://github.com/Nezaemmy/NezaButton/tree/main/NezaButton

1. Move the contents of the `NezaButton` subfolder of the library up one folder level to the root of the repository.
1. Bump the value of the `version` property in the library's `library.properties` metadata file.
1. Create a new [release](https://docs.github.com/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the library repository.

After that, the automated checks will run again. Once those are passing, your library will be automatically added to Library Manager.

Regards, Per

PR #6697, opened 2025-08-01
Removing of `iarduino_*` libraries

cmaglie:
We are removing these libraries because the prefix `iarduino` in the library name and documentation is creating confusion on the owner and source of the library.


stevisco (approved):
Approved

PR #6700, opened 2025-08-01
Revoke registry privileges for github.com/tremaru

cmaglie:
Reference: https://github.com/arduino/library-registry/pull/6697


stevisco (approved):
Approved

PR #6747, opened 2025-08-10
Update repositories.txt

jaikulk14:


per1234 (commented):
Closing as duplicate of https://github.com/arduino/library-registry/pull/6746.

@jaikulk14 please don't submit pointless pull request to this repository. That will only waste your time and the time of the humans who maintain this free open source project.

The Arduino Library Manager Registry contains thousands of libraries from hundreds of authors. Finite resources are available to maintain this resource which is so important to the Arduino community. It is irresponsible to waste those resources.

This is the second time I've had to warn you about irresponsible behavior. If you continue acting in this disrespectful manner your privileges will be revoked and you library removed from the Arduino Library Manager.

PR #6839, opened 2025-08-25
Update repositories.txt

RigelSixSix:
Added flagManager library & examples via github


RigelSixSix:
@ArduinoBot

RigelSixSix:
@ArduinoBot

RigelSixSix:
My apologies - I am new to GitHub - I did not realize a pull request was different than a version update for my code.

Thanks for the human contact - its otherwise quite sterile out there

per1234:
After the initial addition of the library, the indexer job periodically checks each of the libraries in the Arduino Library Manager Registry and picks up any new release/tag that meets the requirements listed here:

PR #6845, opened 2025-08-26
updated: add new repository link for skytraq-navic

cocoisarobo:
Initial release of Skytraq NavIC library

Features:
Configure GNSS constellations (GPS, GLONASS, Galileo, BeiDou, NavIC)
SBAS system configuration
QZSS support
ACK/NACK response handling

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -1,3 +1,4 @@
+https://github.com/cocoisarobo/skytraq-navic
 https://github.com/LordofRobots/LoR_v3
 https://github.com/derdoktor667/DShotRMT
 https://github.com/kenchow96/faster-trig-2040


cocoisarobo:
what? but i've added the maintainer field and the correct url

cocoisarobo:
@ArduinoBot

per1234 (commented):
Hello @cocoisarobo. I see you have fixed the error reported by the bot. However, Library Manager works exclusively from Git tags, so the bot won't be able to see the fix you made until you create a new [release](https://docs.github.com/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the library repository.

https://github.com/cocoisarobo/skytraq-navic/blob/v1.0.0/library.properties

(note that the file did not contain the mandatory `maintainer` property at that tagged point in the revision history)

Regards, Per

cocoisarobo:
@ArduinoBot

cocoisarobo:
@per1234 many thanks!

PR #6852, opened 2025-08-28
Update repositories.txt

arnakazim:
Changing the URL of my library to my new host


per1234 (commented):
Hello @arnakazim. As the bot explained, we only support common Git hosting services like GitHub. So we will not be able to accommodate this request.

Please continue to host a copy of your repository on GitHub for use in the Library Manager distributions. If you prefer to use your alternative Git host for your primary, you can set up a system for automatically mirroring from there to the GitHub repository.

If you do want to fully move the repository away from GitHub, you can use an alternative common Git host such as GitLab. If you do that, please submit another pull request to update the URL to that of the new repository on the supported host.

PR #6944, opened 2025-09-15
a

Droidbane:


Droidbane:
Please help. I added the required properties file everywhere and even created a .json file, but I still continue to get the same error.

per1234 (commented):
Hello @Droidbane. I see you have fixed the error reported by the bot. However, Library Manager works exclusively from Git tags, so the bot won't be able to see the fix you made until you create a new [release](https://docs.github.com/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the library repository.

https://github.com/Droidbane/ESPWiFiManagerByDroidbane/tree/1.0.0

(note that the `library.properties` file was not present in the root of the repository at that tagged point in the revision history)

---

In the future please only create one pull request for each submission.

As was already clearly explained in the comments from the bot, there is no need or point in submitting a new pull request after a problem is detected with your existing PR.

What you must do instead is **CAREFULLY** read the information provided by the bot and then fix the problem. Once you do that, another round of automated checks will be triggered. You can repeat that process in a single pull request as many times as is needed to get it to a fully compliant state. Once your submission passes all the checks it will automatically be accepted by the bot.

You will never fix the problem by submitting the same non-compliant pull request multiple times. That will only waste your time and the time of the humans who maintain this free open source project.

Droidbane:
@ArduinoBot

PR #7046, opened 2025-10-06
bb_temperature

bitbank2:
Please add bb_temperature to the Arduino Library Manager. It's a "universal" I2C sensor library which auto-detects a long list of common sensors. It compiles not only on Arduino, but also on Linux and esp-idf.


bitbank2:
Fixed by doing an initial release @ArduinoBot

PR #7067, opened 2025-10-12
Update repositories.txt

TechTronicsEngineering:
I'm looking to publish my Arduino library (called AutoPlex7) that controls seven segment displays. It works using a timer interrupt and handles multiplexing automatically. This means that AutoPlex7 needs no refresh call and can work with delays in code, providing simplicity and versatility at a whole new level.

Thank you for your time,
Nyjah

PS: sorry if I've submitted this wrong; I'm new to this stuff.


TechTronicsEngineering:
Hello,
I have actually included example sketches. They aren’t visible upon
installation to the Arduino IDE either. I have the folder structure for the
library correct, and library.properties is fine. Not sure what’s wrong with
it.

Thank you for your time,
Nyjah
TechTronics Engineering

PR #7122, opened 2025-10-21
Update repositories.txt

juanjomartinf:
Update repositories.txt


juanjomartinf:
I will fix the problems

PR #7228, opened 2025-11-08
Add SuperCAN repository link

juano2310:
An enhanced Arduino library for CAN bus communication with built-in publish/subscribe protocol.


juano2310:
@ArduinoBot the repository should be public now.

PR #7235, opened 2025-11-11
Providing BLE Serial - NUS

uutzinger:
A library that provides Serial over BLE using the Nordic UART Service.


uutzinger:
Fixed the tag ierror @ArduinoBot

PR #7253, opened 2025-11-16
Add ArduinoDMFLib to registry

marcoratto:
When a message exceeds the maximum payload size of a single packet, it must be fragmented into multiple packets for transmission.
The receiver then reassembles these packets to reconstruct the original message.
Some protocols — such as LoRaWAN and MQTT-SN — require message fragmentation.
This ANSI-C library for Arduino simplifies the process of message fragmentation and reassembly.

### DIFF
--- a/.github/workflows/assets/accesslist.yml
+++ b/.github/workflows/assets/accesslist.yml
@@ -34,6 +34,10 @@
   access: deny
   reference: https://github.com/arduino/library-registry/pull/6514#pullrequestreview-2969201873
 - host: github.com
+  name: Herobrine-pixel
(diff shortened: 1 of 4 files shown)


marcoratto:
When a message exceeds the maximum payload size of a single packet, it must be fragmented into multiple packets for transmission.
The receiver then reassembles these packets to reconstruct the original message.
Some protocols — such as LoRaWAN and MQTT-SN — require message fragmentation.
This ANSI-C library for Arduino simplifies the process of message fragmentation and reassembly.

marcoratto:
I will create new request

PR #7348, opened 2025-12-04
Update repositories.txt

rbuisson-invn:


rbuisson-invn:
@ArduinoBot. Repository set to Public

PR #7369, opened 2025-12-07
Add CRSFforArduino's URL to `repositories.txt`

ZZ-Cat:
This Pull Request adds the URL of my project CRSF for Arduino — A library that brings the Crossfire Protocol (better known as "CRSF") to the Arduino ecosystem.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -1,4 +1,5 @@
 https://github.com/ZZ-Cat/CRSFforArduino
+https://github.com/sava-74/SavaTrig.git
 https://github.com/sava-74/SavaButton.git
 https://github.com/sava-74/SavaTime.git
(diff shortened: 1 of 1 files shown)


ZZ-Cat:
@ArduinoBot Ding, ding! =^/.^=  
I made the necessary changes. Now, let's see what happens.

ZZ-Cat:
> `ERROR: library.properties version value 2025.12.07 is invalid.`

Ah. I know what that is. That's the leading zero on the `PATCH` (IE `07`) field in the version tag.  
I'll fix that up tomorrow, 'cause it's getting late for me.

ZZ-Cat:
@ArduinoBot Pinging you again, as I removed the leading zero in the `PATCH` field and bumped the version number. Let's see what else you got for me.

PR #7425, opened 2025-12-16
Remove "TX07K-TXC Temperature sensor Library"

per1234:
Companion to https://github.com/arduino/library-registry/pull/7424


MatteoPologruto (approved):
Done:
```
2025/12/16 16:11:02 Loaded 8707 libraries from DB
Removing TX07K-TXC Temperature sensor Library
Success!
```

PR #7499, opened 2025-12-30
Remove LF_Alpha repository link from repositories.txt

TanishKunthe:
Removed a repository link from the list.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -4887,6 +4887,7 @@ https://github.com/MSZ98/serial-readline
 https://github.com/mt-elektrik/MteCore
 https://github.com/mtsurumaki/PwmSin
 https://github.com/mudmin/ESPCanary
+https://github.com/mussacharles60/StepperShiftRegister74HC595
 https://github.com/musicplayer0/solar2lunar
 https://github.com/muwerk/mufonts
 https://github.com/muwerk/munet


per1234 (commented):
Hello @TanishKunthe. Please add a comment here on the pull request thread that provides a detailed explanation for why you wish to remove this library.

We do remove libraries on request, but I find it strange that you would request a removal only a few days after submitting the library. Additionally, you submitted another completely pointless pull request after that https://github.com/arduino/library-registry/pull/7482. This makes me think that your true goal is not actually the permanent removal of the library, but instead that you plan to remove and then re-add the library in hopes that doing so will accomplish some other goal. That is not allowed and would not be productive anyway. We only allow removal requests when the intent is the permanent removal of the library.

TanishKunthe:
Hello @per1234,

I apologies for my mistakes and Thank you for the follow-up. To be transparent, you are correct about my intent, and I apologize for the confusion caused by the previous PR #7482.

My goal is to ensure the library launches cleanly with Version 1.0.0. Due to the naming errors in my initial submission, I realized that if I simply update the existing entry, I would likely be forced to bump the version to 1.0.1 immediately. I strongly prefer to avoid this so that the initial official release is properly versioned as 1.0.0.

I understand the policy against removing and re-adding libraries. However, since this was a mistake during the initial registration and the library hasn't been widely used yet, I was hoping to get a "clean slate" to register the corrected version.

Is it possible to proceed with the removal in this specific context, or is there a better way to correct the name while preserving the 1.0.0 tag?

PR #7537, opened 2026-01-08
Remove FastLED_NeoPixel

dmadison:
This library is intentionally being removed from the registry as its dependency is no longer functional.


dmadison:
Thank you @MatteoPologruto!

PR #7545, opened 2026-01-09
Fix repo URLs to reference user WolodiaM instead of Wolodia-M

WolodiaM:
I have changed my account name to WolodiaM (because I deleted my other account, which was on different email (and this account was unused)). For now it works because GitHub aliases repos between users, but anyone can register Wolodia-M now and if this person creates repo with same name as my library it will be redirected to that person's repo. If some proofs needed that it is actually me who owns repos, which urls were changed, I can try to provide them.


WolodiaM:
Fixed missing LICENSE files in flagsapi and timersapi libraries. @ArduinoBot

WolodiaM:
Actually added LICENSE to timersapi. I think I forgot to commit previous time. @ArduinoBot

PR #7695, opened 2026-02-07
Menambahkan Library Arduino untuk WhatsApp API Gateway

Sebuahhobi:
Menambahkan library Arduino untuk menghubungkan Arduino dengan WhatsApp API Gateway. Library ini memungkinkan pengiriman pesan WhatsApp langsung dari perangkat Arduino.
Dukungan ESP32/ESP8266

PR #7699, opened 2026-02-08
Add Arduino Library for WhatsApp API Gateway with ESP32/ESP8266

Sebuahhobi:
This library allows Arduino boards to send messages using the WhatsApp API Gateway, enabling seamless integration of WhatsApp messaging capabilities in various IoT projects. It supports ESP32 and ESP8266 boards, making it ideal for IoT applications that require reliable messaging.

Key features:

Send WhatsApp messages from Arduino via the WhatsApp API Gateway.

Compatible with ESP32 and ESP8266 for Wi-Fi connectivity.

Ideal for applications such as remote monitoring, alerts, and automation.

Easy integration with existing Arduino projects.

The library includes example sketches to help users get started with sending messages to WhatsApp groups or individual contacts.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -8932,3 +8932,4 @@ https://github.com/sapteinkabeltann/lua511-esp32
 https://github.com/louischuang/curiejet-p760-arduino-lib.git
 https://github.com/bkwoka/SemVerChecker
 https://github.com/alxprgs/RTC_RDP_LIB_L298N
+https://github.com/ivica3730k/MSF-Time-Lib


per1234 (commented):
@Sebuahhobi this pull request has the same problem I asked you to fix in your other pull request:

https://github.com/arduino/library-registry/pull/7695#pullrequestreview-3767819364

It is becoming very time consuming and tedious reviewing all these invalid pull requests from you. I need you to make a much stronger effort in your activity in this repository.

You started off badly by ignoring the instructions we provide and instead improvising on how you thought a submission might be accomplished.

You then submitted three duplicate pull requests with the same problem as the first even though if you had taken even a few seconds to review the changes that would be proposed by the pull request, it would have been clear to you that they were invalid.

Our goal is to enable everyone to create Arduino projects in the manner that suits them. However, the situation is different when it comes to distributing libraries to the entire Arduino community via the Arduino Library Manager. We expect a high level of competence from the library maintainers who use the registry. If you are not able to meet that standard, then you will not be able to properly maintain a library that is suitable for use by the community.

So it is time for you to either step up your game, or else step away. If you continue to waste my time, I will revoke your Library Registry privileges.

PR #7704, opened 2026-02-08
Update repositories.txt

RobTillaart:
Arduino library for DAC8562 SPI 16 bit dual channel DAC and compatibles.


RobTillaart:
@ArduinoBot  oops

PR #7752, opened 2026-02-17
Remove my library URL from repositories.txt

Embedded-Petal:
This pull request removes my library URL from repositories.txt as I no longer want it included.


Embedded-Petal:
I am the author of this library and I would like to remove it from the Arduino Library Manager.

Reason:
This library is no longer maintained / I no longer want to distribute it via the Arduino Library Manager.

I have intentionally removed the repository URL from repositories.txt.

Please remove this library from the registry.

PR #7756, opened 2026-02-17
Fixed bug in MoIRA v1.0.1

Ada-Rhodes:
There was an error in how the dependency was listed


Ada-Rhodes:
I think I followed the wrong process for updating the library, I believe I found the correct documentation now here (https://github.com/arduino/library-registry/blob/main/FAQ.md#how-can-i-publish-a-new-release-once-my-library-is-in-the-list:~:text=Updates-,How%20can%20I%20publish%20a%20new%20release%20once%20my%20library%20is%20in%20the%20list%3F,-Make%20sure%20the) 

Sorry about that

PR #7802, opened 2026-02-24
add voyager-ota-client 

mediocre9:


mediocre9:
@ArduinoBot please review and check again....

PR #7812, opened 2026-02-25
Add Grobot_Animations library

tanmaywankar:
Adding my Grobot_Animations library to the registry. It is compliant with the 1.5 specification and tested on ESP32.


tanmaywankar:
@ArduinoBot

PR #7876, opened 2026-03-09
Update repositories.txt

beastbroak30:


beastbroak30:
@ArduinoBot please re-check. Release v1.2.0 has been created with the updated library.

PR #7884, opened 2026-03-10
Add new repository link for boost-embedded-190

jcelerier:
This allows to use a recent version of boost ; many libraries work fine for instance on ESP32, Teensy, etc. as most things are header-only


jcelerier:
@ArduinoBot

PR #7860, opened 2026-03-05
Add Forgetfulino library

IamTheVector:
This PR adds the Forgetfulino library by IamTheVector to the Arduino Library Manager registry.

PR #7932, opened 2026-03-16
Add Forgetfulino subdirectory URL

IamTheVector:
This PR adds the URL below to repositories.txt as requested:

- https://github.com/IamTheVector/Forgetfulino/tree/main/Forgetfulino-main

Note: This is a repository subdirectory URL (tree).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -9049,3 +9049,4 @@ https://github.com/cubicleguy/xv7021_arduino_spi
 https://github.com/Kiwistron/KiwisIoT
 https://github.com/EdwinKestler/SIM800_MQTT.git
 https://github.com/IamTheVector/Forgetfulino
+https://github.com/IamTheVector/Forgetfulino/tree/main/Forgetfulino-main


IamTheVector:
> Hello @IamTheVector. The `repositories.txt` file is a list of Git repository URLs. It is required that the library be located in the root of the repository. There is no support for specifying a subfolder in the repository.
> 
> You already successfully registered the library some time ago via #7860, so I'm not sure what you were attempting to accomplish here. If you were attempting to accomplish something other than adding an invalid and redundant URL to the list, you can provide a detailed explanation in a comment here on this pull request thread and I'll provide assistance.

Hello,

Sorry for the confusion and the invalid/redundant URL.

What happened:
I tried to add a repository URL that pointed to a subfolder because my repo contains both the Arduino library and a large (~100 MB) IDE extension project (the tool that uses the library). I was hoping to reference only the library part via a subfolder.

How I fixed it:
I’ve now separated the two projects into two Git repositories: the library lives in its own repo (at the root), and the extension in another. So the Library Manager only needs the library repo, with the library at the root, which matches the requirement. I’m not trying to add a new or different URL for the same library; the registration from #7860 remains the correct one.

Summary:
I misunderstood the “library at root only” rule and tried to work around the big extension folder by pointing to a subfolder to avoid usesrs to download unused build data. The proper solution was to split library and extension into two repos and keep the existing library registration. Sorry again for the noise on the list.

Thanks for the clarification and sorry for the confusion.

PR #7977, opened 2026-03-22
Add Relay-Controller library

udfvirus:
This is a pull request for a flexible OOP-based library for controlling relays via GPIO pins or Serial commands.


udfvirus:
@ArduinoBot fixed: ❌ ERROR: The repository has no tags. You need to create a [release](https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) that matches the version value in the library's library.properties file.

PR #8012, opened 2026-03-29
Update repositories.txt

mumanchu:


mumanchu:
@ArduinoBot says "the repository ash no tags or releases". Added release v1.0.0.

PR #8050, opened 2026-04-04
Add new repository link to repositories.txt

m0spak:


m0spak:
@ArduinoBot The library AdditiveHW is listed in the Arduino Library Registry [repositories.txt](https://raw.githubusercontent.com/arduino/library-registry/refs/heads/main/repositories.txt), but is currently unavailable at the corresponding Arduino documentation URL: https://docs.arduino.cc/libraries/additivehw/ .

per1234:
Hi @m0spak. I apologize for the problem with the [**Arduino Library Reference**](https://docs.arduino.cc/libraries/). Although the **Arduino Library Reference** pages are generated from data obtained from the [Library Manager index](http://downloads.arduino.cc/libraries/library_index.json), which is in turn generated from the Arduino Library Registry data, the reference generation system is completely separate (both in code base and maintainership) from this repository, and from any of the other associated Library Manager infrastructure. So the maintainers of this repository are unable to assist with issues related to the **Arduino Library Reference**.

Unfortunately there isn't a public issue tracker for the **Arduino Library Reference** generation system and its infrastructure, but you can submit a ticket via the contact form here:

https://www.arduino.cc/en/contact-us/

That will be the most efficient way to bring the problem to the attention of those who are responsible for the **Arduino Library Reference** generation system.

---

I am happy to report that the system I maintain is working correctly, as we can see that your library is available for installation from Arduino IDE (as well as other official Arduino development tools):

image

PR #8150, opened 2026-04-16
Added TFTGraph as a library

SeventyFourProductions:


SeventyFourProductions:
@ArduinoBot I think I fixed it, added release 1.0.0 now.

PR #8181, opened 2026-04-20
Add foodini/arduino/ms5611_spi

foodini:
This is an spi-based driver for a compact, surface-mounted altitude/pressure/temperature sensor.  The interface is a simple command-and-tick style process where you just call tick until it tells you that it is done.


foodini:
@ArduinoBot Newline issue fixed, tag has been created.

PR #8193, opened 2026-04-23
add soldered de2120 barcode scanner

franFodor:


franFodor:
@ArduinoBot release tag has been made

PR #8192, opened 2026-04-23
Update repositories.txt

hestialabs:
The library was previously registered under:
https://github.com/hestia1abs/hxtp-micro

This repository has been deleted and replaced by:
https://github.com/hestialabs/hxtp-micro

The new repository is already registered and should remain.

Request:
Please remove the obsolete entry to avoid duplication and confusion.

PR #8205, opened 2026-04-25
Remove "hxtp-micro"

per1234:
Companion to https://github.com/arduino/library-registry/pull/8192

### DIFF
--- a/registry.txt
+++ b/registry.txt
@@ -9060,7 +9060,6 @@ https://github.com/DIYables/DIYables_OLED_SSD1309.git|Contributed|DIYables_OLED_
 https://github.com/ASNProject/Panzer_Robotics_Library.git|Contributed|Panzer_Robotics_Library
 https://github.com/roboticist-blip/LoRa32T.git|Contributed|LoRa32T
 https://github.com/dhruvkotian/em2m-esp32-modbus.git|Contributed|EM2M-ESP32-Driver
-https://github.com/hestia1abs/hxtp-micro.git|Contributed|hxtp-micro
 https://github.com/GyverLibs/GyverMotor2.git|Contributed|GyverMotor2
 https://github.com/GyverLibs/GyverKey.git|Contributed|GyverKey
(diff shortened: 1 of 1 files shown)


sddion:
@ArduinoBot

hestialabs:
Can i get a Eta @per1234

per1234:
Hi @hestialabs. Unlike submissions, this type of request is still handled manually. The Library Manager index maintainer occasionally comes through to clear all the outstanding requests and closes those issues/PRs when that is done.

So the delay before the library is removed is dependent on what point in the cycle your request happened to fall. You can follow the issue/PRs to receive a notification from GitHub when it is closed. After that, the change may still not take effect immediately in Library Manager on the machines of individual users. The reason for this is that the index that populates Library Manager is cached locally for the sake of efficiency, and it may take some hours before the update to the index propagates through the cache.

We'll surely get to it soon, please be patient.

---

I'm not clear on why you seem to feel this is so urgent. Your replacement library has already been registered, so there is nothing preventing users from using that one right now. This makes me suspect that maybe you have some other goal that you haven't communicated to me, and that you believe that the removal of the library will somehow result in that goal being accomplished or unblocked. It is possible that will be the case, but not guaranteed. So if this is something you are anxious about, it would behoove you to clearly communicate what you are actually trying to accomplish. That will allow me to effectively assist you in accomplishing your goals.

sddion:
@per1234 

The older library is no longer maintained and can cause confusion for users. 

I’m currently building a startup https://docs.hestialabs.in around this ecosystem, so clarity and correctness in the registry are important for adoption.


This isn’t about any hidden goal—just ensuring users don’t end up using something obsolete. I’d appreciate prioritizing the removal when possible.

MatteoPologruto (approved):
Done:
```
2026/04/30 15:47:35 Loaded 9312 libraries from DB
Removing hxtp-micro
warning: Library Git clone folder /home/ubuntu/workspace/gitclones/github.com/hestia1abs/hxtp-micro not present
Success!
```

PR #8291, opened 2026-05-09
Revert "Update repositories.txt (#8290)"

ulywae:
This reverts commit adf2094b33432b3039b9c001851ae13776fa096e.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -9383,4 +9383,3 @@ https://github.com/sauloverissimo/midi2_cpp
 https://github.com/RobotBuzzard/ClearCoreWatchdog
 https://github.com/soosp/SimcomA76xx
 https://github.com/K3vin135/iot-azure-bridge
-https://github.com/ulywae/NeuStorage


per1234 (commented):
Hello @ulywae. Please provide a detailed explanation of the purpose of this change.

We do allow library maintainers to submit requests for the removal of libraries from the registry. However, this must only be done when the intention is to permanently remove the library from the registry. You submitted this pull request only hours after registering the library (https://github.com/arduino/library-registry/pull/8290), so I'm very doubtful that it is truly your intent to permanently remove the library from the registry. This makes me think that you are trying to accomplish some other goal in a hacky manner. That is not going to be an effective or efficient way to accomplish your goal, and is also likely to waste the limited resources we have to allocate to maintaining this resource that is so important to the entire Arduino community.

So I need you to make the effort to clearly communicate with me. I am here to help you, but I can only do so if I understand what you need help with.

ulywae:
Please I want to delete the NeuStorage library because there is a mistake
in its naming.

Pada tanggal Sab, 9 Mei 2026 23.21, Per Tillisch ***@***.***>
menulis:

per1234:
@ulywae that is not an effective way to correct the name of a library. If you want to change the library name, then follow the correct procedure:

https://github.com/arduino/library-registry/blob/main/maintenance-requests.md#update-library-name:~:text=accommodating%20your%20request.-,Update%20library%20name,-For%20the%20sake

PR #8332, opened 2026-05-11
Update URL of "Universal Blink"

per1234:
Companion to https://github.com/arduino/library-registry/pull/8323


per1234 (commented):
Superseded by https://github.com/arduino/library-registry/pull/8404

Issue #8314, opened 2026-05-10
Library name change request


hestialabs:
### Library repository URL

https://github.com/hestialabs/hxtp-micro

hestialabs:
<img width="878" height="197" alt="Image" src="https://github.com/user-attachments/assets/33cd6c47-98ba-4f9b-98a6-a30508864d1e" />

hestialabs:
@per1234 Hi, I’m facing an issue with my Arduino library registry entry.
My library is correctly named `hxtp-micro` across the GitHub repo, `library.properties`, and releases, but on `downloads.arduino.cc/libraries/logs/github` it appears as `HXTP embedded SDK`, which I never configured or submitted.

Could you please help me understand where this incorrect name is coming from and how to fix it?


I’m also facing another issue related to old cached library metadata.

When searching on Google and Arduino library indexes, my project still points to an old library/version and old repository history that no longer exists. I already removed/renamed the old GitHub repository and updated the library name for the registry, but outdated entries are still appearing.

Example:
https://www.arduinolibraries.info/libraries/hxtp-micro

How can I fully remove or refresh the old cached registry data?

per1234:
> Could you please help me understand where this incorrect name is coming from

The library name is defined by the `name` property in the library's `library.properties` metadata file. When you register a library here, it is registered under the name defined by the value of that property at the time of the latest Git tag. I see that there is no evidence for this in the revision history of your repository. However, I also see you have removed all the older tags from the repository. You have been up to a lot of other shenanigans related to this library (https://github.com/arduino/library-registry/pull/8192#pullrequestreview-4163856349), so I think it is likely that the revision history we see now is quite different from what the submission system was working from at the time of the registration (https://github.com/arduino/library-registry/commit/dafc58f44138051959f6db998cbd99a3af15c3f6).

> how to fix it?

By submitting a rename request, as you have done here.

> Example:
>https://www.arduinolibraries.info/libraries/hxtp-micro
>
>How can I fully remove or refresh the old cached registry data?

Note that arduinolibraries.info is a 3rd party website. It is generated from the data of the [Arduino Library Manager index](http://downloads.arduino.cc/libraries/library_index.json), but the Arduino company doesn't have any involvement in the maintenance and support of that website.

Note the information in the footer of the site:

> The content on this site was last updated: 2026-04-16 04:03:49 +0100

[...]

PR #8333, opened 2026-05-11
Update name of "hxtp-micro"

per1234:
Resolves https://github.com/arduino/library-registry/issues/8314


MatteoPologruto (approved):
```
2026/05/25 16:41:39 Loaded 9442 libraries from DB
error: Library name HXTP Embedded SDK not found
```
The library is not present inside the index. I'm approving and merging the rename request.

PR #8542, opened 2026-06-14
Remove libraries: Virtuino, virtuinoESP, virtuino_stm32

iliaslamprou:
Removing this library as it is no longer maintained.

Repository: https://github.com/iliaslamprou/virtuino
Repository: https://github.com/iliaslamprou/virtuinoESP
Repository: https://github.com/iliaslamprou/virtuino_stm32

PR #8565, opened 2026-06-18
Remove "Virtuino"

per1234:
Companion to https://github.com/arduino/library-registry/pull/8542


MatteoPologruto (approved):
Done:
```
2026/07/03 15:54:14 Loaded 9631 libraries from DB
Removing Virtuino
Success!
```

PR #8636, opened 2026-06-26
CoVACIEL-ESP32

rgot-org:
Bibliothèque Arduino pour le véhicule autonome CoVACIEL (ESP32).


per1234 (commented):
Closing as duplicate of https://github.com/arduino/library-registry/pull/8144

@rgot-org please don't submit pointless pull requests to this repository. The Arduino Library Manager Registry contains thousands of libraries from hundreds of authors. Finite resources are available to maintain this resource which is so important to the Arduino community. Pointless pull requests waste those resources and so this is irresponsible behavior.

---

rgot-org:
I have made the CoVACIEL-ESP32 library available for Arduino.
It appears as registered when I visit https://docs.arduino.cc/libraries/covaciel/
The displayed version is 0.0.3, however the latest version on GitHub is 1.0.3.
But when I go to the Arduino Library Manager, it does not appear in the search engine.
Can you help me?

per1234:
You can see the cause of the problem here in the library's dedicated indexer logs page:

https://downloads.arduino.cc/libraries/logs/github.com/rgot-org/CoVACIEL-ESP32/

```text
2026/06/27 09:30:15 Checking out tag: 1.0.3
2026/06/27 09:30:15 Release CoVACIEL-ESP32:1.0.3 has wrong library name, should be CoVACIEL
```

The library had the name "CoVACIEL" at the time of registration:

https://github.com/arduino/library-registry/commit/039f778cecae935548f9c54ceb22afe3dfeb2cf4

Since that time, you changed the name to "CoVACIEL-ESP32":

https://github.com/rgot-org/CoVACIEL-ESP32/commit/d872eb0256acfee9bb6b694a08858dbec665f854

The reason for the name change restriction is that the name is the unique identifier for the library, meaning that any changes to the identifier might cause problems to users who rely on the established identifier in command line operations or for other library developers who have defined it as one of their dependencies.

https://github.com/arduino/library-registry/blob/main/maintenance-requests.md#update-library-name

PR #8708, opened 2026-07-07
Update repositories.txt

nthnsdev:


per1234 (commented):
The Arduino Library Registry is used to distribute libraries to the Arduino community. Libraries should only be registered if you believe they will be of value to Arduino community members.

The library you are proposing to add doesn't have any content, so it is objectively of no value. For this reason, I will decline your submission.

@nthnsdev in the future, please only submit libraries if you truly believe they will be of value to Arduino community members. Submitting useless libraries is irresponsible.

The Arduino Library Manager Registry contains thousands of libraries from hundreds of authors. Finite resources are available to maintain this resource which is so important to the Arduino community. It is irresponsible to waste those resources.

PR #8808, opened 2026-07-24
Add new repository link for VQF-for-Arduino

LZQS1:


LZQS1:
@ArduinoBot

per1234 (commented):
Hello @LZQS1. I see you have fixed the error reported by the bot. However, Library Manager works exclusively from Git tags, so the bot won't be able to see the fix you made until you create a new [release](https://docs.github.com/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the library repository.

If you look at the contents of the repository as they were at the time of the latest tagged revision, you can see that the version of the library the bot is checking still has the problem:

https://github.com/LZQS1/VQF-for-Arduino/blob/v1.0.0/library.properties#L8

```text
url=
```

Regards, Per

LZQS1:
@ArduinoBot

PR #8965, opened 2026-08-19
Update repositories.txt

NachtRaveVL:
Adding three more of my libraries into the registry

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -5215,7 +5215,6 @@ https://github.com/nacademyth/RoboFi
 https://github.com/NachtRaveVL/BY8X01-16P-Arduino
 https://github.com/NachtRaveVL/Lepton-FLiR-Arduino
 https://github.com/NachtRaveVL/PCA9685-Arduino
-https://github.com/NachtRaveVL/SuperFX3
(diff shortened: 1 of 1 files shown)


per1234 (commented):
Hello @NachtRaveVL. I see you have fixed the error reported by the bot when it checked the "SuperFX3" library. However, Library Manager works exclusively from Git tags, so the bot won't be able to see the fix you made until you create a new [release](https://docs.github.com/github/administering-a-repository/releasing-projects-on-github/managing-releases-in-a-repository) or [tag](https://git-scm.com/docs/git-tag) in the SuperFX3 library repository.

If you look at the contents of the repository as they were at the time of the latest tagged revision, you can see that the version of the library the bot is checking still has the problem:

https://github.com/NachtRaveVL/SuperFX3/tree/0.9.0

(note that the `library.properties` metadata file was missing at the tagged point in the revision history)

Regards, Per

NachtRaveVL:
try again @ArduinoBot - ive updated the files.

NachtRaveVL:
@ArduinoBot okay try again once again

NachtRaveVL:
alright @ArduinoBot we're just going to remove that one for now until we get it fixed

PR #9021, opened 2026-09-01
Add new libraries to repositories .txt, DVI for RP2040

obviousfancy:
Adds support for a DVI output library covering both RP2040 (PIO-based) and RP2350 (HSTX-based) implementations.


obviousfancy:
@ArduinoBot the libraries are public right now

PR #9039, opened 2026-09-05
Rename PedroRobot repository to PedroFirmware

almtzr:
Hi,

I would like to rename my repository library.
Sorry I didn't know if it's possible to rename my repository library by a pull request.

Waiting for you for the correct process if it is needed.

Thanks

PR #9041, opened 2026-09-05
Update name and URL of "PedroFirmware"

per1234:

- Change URL from `https://github.com/almtzr/PedroRobot.git` to `https://github.com/almtzr/PedroFirmware.git` (companion to https://github.com/arduino/library-registry/pull/9039).
- Change name from `PedroRobot` to `PedroFirmware`

Since the name change operation on the database is actually a removal followed by automated re-indexing of the updated registry entry on the next job run, the URL update will occur as a matter of course. For this reason, the only operation required from the backend maintainer is a standard name change procedure.


MatteoPologruto (approved):
Done:
```
2026/09/14 15:07:22 Loaded 9969 libraries from DB
Removing PedroRobot
Success!
```

PR #9071, opened 2026-09-09
remove my lib RingBufferStream

atesin:
please remove my lib RingBufferStream because i want to relicense and fix some bugs... i was researching and realized wtfpl could be risky in some cases so want to put under a more serious license, and also take the chance to fix some bugs i found... so i want to delete and rollback everything, thanks


atesin:
please tell me i can remove the lib, then redo it, then publish again... if not then this removal request would have no sense :)

per1234 (commented):
Hi @atesin. We only allow library removals when the intent is to permanently remove the library from the repository.

In this case where you only want to change the content of the library, you must instead simply make a new release to distribute those changes to the users:

PR #9077, opened 2026-09-10
Add new repository link for FixedString

Monsterovich:
Origin: https://gitlab.com/Monsterovich/fixedstring

FixedString is a header-only C++ library providing FixedString<N> fixed-size string class with zero heap allocation, designed for memory-constrained embedded systems.

### DIFF
--- a/repositories.txt
+++ b/repositories.txt
@@ -3190,7 +3190,6 @@ https://github.com/FastLED/FastLED
 https://github.com/FatBeard/vbus-arduino-library
 https://github.com/fatihaslamaci/TimerFa
 https://github.com/fatkhrus/TeslaCloud
-https://github.com/fatlab101/FixedString
 https://github.com/fatpat/arduino-dra818
 https://github.com/fbiego/chronos-esp32
 https://github.com/fbiego/CST816S
--- a/repositories.txt
+++ b/repositories.txt
@@ -9985,3 +9985,4 @@ https://github.com/aschet/tinyolfarve
 https://github.com/panjingwei1945/PWM_Pulse
 https://github.com/thecaptainexe/OLEDAnimator
 https://github.com/ReP64/CH462x
+https://gitlab.com/Monsterovich/fixedstring


per1234 (commented):
Hello @Monsterovich. Thanks for your interest in contributing to the Arduino Library Manager.

>A problem was found with your submission https://gitlab.com/Monsterovich/fixedstring
>
>❌ ERROR: The repository has no tags. You need to create a release or tag that matches the version value in the library's library.properties file.

1. Create a new [release](https://docs.gitlab.com/user/project/releases/#create-a-release-in-the-releases-page) or [tag](https://git-scm.com/docs/git-tag) in the library repository.

After that, the automated checks will run again. Once those are passing, your library will be automatically added to Library Manager.

Regards, Per

Monsterovich:
@per1234 Thank you. I've already added the tags.

Monsterovich:
@ArduinoBot I've fixed the tags. Check again.

Monsterovich:
Now there is one more `FixedString` implementation, and both are worse than mine. Heh heh.

https://github.com/fatlab101/FixedString
https://github.com/toomasz/FixedString

I don't know what to do. To be honest, I'm too lazy to deal with this anymore.

Monsterovich:
I asked Claude to do analysis.

<html><head></head><body><h1>Comparing Three <code>FixedString</code> Implementations</h1>
<p>All three projects solve the same core problem — a <code>std::string</code>-like, non-heap-allocating string type for embedded/Arduino targets, sized at compile time via a template parameter (<code>FixedString&lt;N&gt;</code>). Beyond that, they diverge quite a bit in design philosophy, API shape, and how much "extra" they build on top.</p>
<h2>1. <code>fatlab101/FixedString</code></h2>
<ul>
<li><strong>Storage model</strong>: single class <code>FixedString&lt;c_storage_size = 64&gt;</code>. The template parameter is the <em>total byte footprint</em> of the object (must be 4–256 and a multiple of 4), not the usable capacity — usable capacity is <code>c_storage_size - 2</code>. Length is cached in a single <code>unsigned char</code>, so max string length is hard-capped at 255 regardless of <code>c_storage_size</code>.</li>
<li><strong>API style</strong>: mirrors Arduino's <code>String</code> class closely (<code>concat</code>, <code>compareTo</code>, <code>equalsIgnoreCase</code>, <code>indexOf</code>, <code>lastIndexOf</code>, <code>toCharArray</code>, <code>startsWith</code>/<code>endsWith</code>, <code>toInt</code>/<code>toFloat</code>, <code>replace</code>, <code>trim</code>, <code>toUpperCase</code>/<code>toLowerCase</code>) rather than <code>std::string</code>.</li>
<li><strong>Numeric support is a first-class citizen</strong>: dedicated constructors/<code>assign()</code> overloads for <code>char</code>, <code>int</code>, <code>long</code>, unsigned variants, <code>float</code>, <code>double</code>, with explicit <code>Radix</code> (base2/8/10/16) support, and a non-Arduino path that uses C++17 <code>std::to_chars</code> for conversions.</li>
<li><strong>Overflow handling</strong>: most mutating operations return <code>bool</code>; on overflow it truncates safely and calls a <code>notify_overrun()</code> hook that just <code>printf</code>s a warning by default (no exception, no global flag).</li>
<li><strong>Flash-string (PROGMEM) support</strong>: has explicit macros (<code>DEFINE_PSTR</code>/<code>GET_PSTR</code>) and constructors/assign paths for <code>__FlashStringHelper*</code>, and a <code>format()</code> overload for flash-stored format strings — clearly built with real AVR flash constraints in mind.</li>
<li><strong>No dynamic dispatch, <code>final</code> class</strong>, no inheritance hierarchy — one self-contained template.</li>
<li>Requires <strong>C++17</strong> for the non-Arduino ("testing") build path, since it leans on <code>&lt;charconv&gt;</code>.</li>
</ul>
<h2>2. <code>toomasz/FixedString</code></h2>
<ul>
<li><strong>Storage model</strong>: <code>FixedString&lt;N&gt;</code> (via <code>SimpleString&lt;N&gt;</code>-style aliasing) where <code>N</code> is closer to the usable size; overhead is documented as a fixed 5 bytes per instance (2 for length, 2 for capacity, 1 for the null terminator), so e.g. <code>FixedString&lt;20&gt;</code> occupies 25 bytes.</li>
<li><strong>API style</strong>: smaller and simpler than the other two — <code>length()</code>, <code>c_str()</code>, <code>append()</code>, <code>appendFormat()</code> (printf-style), operator <code>+=</code>, and little else. The README even lists missing features like <code>replace</code>/<code>trim</code> as a "Todo" for contributors.</li>
<li><strong>Overflow handling is the most distinctive part of the design</strong>: instead of per-call truncation flags or exceptions, an overrun sets a single <strong>global</strong> variable <code>FixedString_OverflowDetected</code>. The intended usage pattern is to check that flag once in the main loop and halt/fault the whole program — a "fail loud, fail global" philosophy suited to a small sketch where any overflow is a bug you want to catch immediately, not something to recover from per-call.</li>
<li><strong>Positioning</strong>: explicitly framed as a <code>String</code> replacement to avoid Arduino/ESP8266 heap fragmentation, with the README citing the classic "evils of Arduino String" articles as motivation. It's the most minimal, "get me a safe fixed buffer + concat" tool of the three.</li>
<li>Smallest codebase, least actively developed-looking API surface, but also the simplest mental model.</li>
</ul>
<h2>3. The uploaded <code>FixedString.h</code> ("Monsterovich")</h2>
<p>This is by far the most elaborate of the three, structured more like a hand-rolled subset of <code>std::string</code> than an Arduino <code>String</code> clone.</p>
<ul>
<li><strong>Two-layer design</strong>: a non-owning <code>FixedStringBase</code> (holds a <code>char*</code> + capacity, all logic lives here) and a thin <code>FixedString&lt;N&gt;</code> that owns a <code>char buffer_[N+1]</code> and constructs the base against it. This lets you write functions that accept <code>const FixedStringBase&amp;</code> regardless of the caller's <code>N</code> — something neither of the other two designs supports (they need same-<code>N</code> or templated-<code>M</code> overloads to interoperate).</li>
<li><strong><code>std::string</code>-shaped API</strong>: <code>begin()/end()</code>, iterators, <code>substr</code>, <code>insert</code>, <code>erase</code>, <code>replace</code>, <code>find</code>/<code>rfind</code>/<code>find_first_of</code>/<code>find_last_not_of</code> family, <code>compare</code>, relational operators, <code>operator&lt;&lt;</code> for <code>std::ostream</code> — essentially the STL string surface — <em>plus</em> Arduino-compatible aliases (<code>charAt</code>, <code>startsWith</code>, <code>substring(begin,end)</code>) for drop-in familiarity.</li>
<li><strong>Portability handling is very deliberate</strong>: guards around <code>&lt;initializer_list&gt;</code>, <code>&lt;stdexcept&gt;</code>, and <code>&lt;ostream&gt;</code> for AVR toolchains that lack libstdc++, a configurable <code>FIXEDSTRING_NO_EXCEPTIONS</code>/<code>FIXEDSTRING_ON_ERROR</code> hook system (call a diagnostic hook, then optionally throw), separate from the other two libraries' simpler approaches (printf warning vs. global flag).</li>
<li><strong>Self-aliasing safety</strong>: <code>insert</code>/<code>replace</code> explicitly detect when the source pointer aliases the object's own buffer (via <code>aliases_buffer()</code>, comparing <code>uintptr_t</code>s to avoid pointer-comparison UB) and take a rotate-based or insert-then-erase path to stay correct — a subtlety neither other implementation appears to address (they don't document overlap safety for self-referential <code>concat</code>/<code>replace</code> calls).</li>
<li><strong>Truncation semantics as a queryable, non-fatal state</strong>: <code>is_truncated()</code> is a plain per-object flag (no global, no exception required) that composes with the operation return values — a middle ground between fatlab101's per-call <code>bool</code> and toomasz's process-wide flag.</li>
<li><strong>Extra composite type</strong>: <code>CombinedString&lt;N, MaxSegments&gt;</code> — a zero-copy, lazily-built concatenation of multiple <code>FixedStringBase</code> segments (pointers only, no copying until <code>.build()</code>/first access). Nothing comparable exists in the other two.</li>
<li><code>format()</code> is a variadic-template <code>snprintf</code> wrapper (not C varargs), giving compile-time type checking of arguments — arguably safer than the other two libraries' <code>va_list</code>-based <code>format()</code>/<code>appendFormat()</code>.</li>
<li><strong>PROGMEM/<code>__FlashStringHelper</code> support was deliberately left out</strong>, not overlooked: the object still keeps its own RAM copy of the string either way, so <code>F("...")</code> would only avoid duplicating the literal in RAM at startup — a modest saving, not a saving on the buffer itself. Meanwhile <code>__FlashStringHelper</code> is AVR/Arduino-specific and would require <code>#ifdef ARDUINO</code>, <code>memcpy_P</code>/<code>strlen_P</code>, and new overloads scattered across the whole API (constructor, <code>assign</code>, <code>append</code>, <code>insert</code>, <code>replace</code>, <code>compare</code>, ...) — a large surface-area increase for a small payoff. A narrower, cheaper version of this (support only in the constructor/<code>assign</code>/<code>append</code> from <code>const __FlashStringHelper*</code>, the most common use case) was considered a reasonable compromise but wasn't implemented; it would be documented as a known limitation instead.</li>
</ul>
<h2>Summary Table</h2>

Aspect | fatlab101 | toomasz | Uploaded ("Monsterovich")
-- | -- | -- | --
API model | Arduino String-like | Minimal, custom | std::string-like + Arduino aliases
Overflow signal | Per-call bool + overrun hook | Global bool flag | Per-object is_truncated() flag (+ optional exception)
Numeric conversions | Extensive (Radix, float/double, to_chars) | Via appendFormat only | Via format() only
PROGMEM/flash support | Yes, explicit macros | No | Deliberately omitted (low value vs. API-wide cost)
Cross-N interop | Templated copy ctor | Not evident from README | FixedStringBase& works across any N
Self-aliasing safety | Not documented | Not documented | Explicit, handled
Extra composite types | None | None | CombinedString (lazy multi-segment concat)
Codebase size/complexity | Medium | Small | Large
License | LGPL-2.1 | MIT | MIT


<h2>Bottom line</h2>
<ul>
<li><strong>toomasz/FixedString</strong> is the leanest option — pick it if you want the smallest possible footprint and are fine with a single global overflow flag and a bare-bones API.</li>
<li><strong>fatlab101/FixedString</strong> sits in the middle, optimized specifically for Arduino idioms (drop-in <code>String</code> replacement) and numeric/flash-string conversion work.</li>
<li><strong>The uploaded implementation</strong> is the most feature-complete and defensively engineered of the three (STL-like surface, aliasing safety, per-object truncation tracking, lazy concatenation), at the cost of being the largest and most complex header to audit and compile.</li>
</ul></body></html>

per1234 (commented):
@Monsterovich you fumbled your rebase. In addition to correctly adding the `https://gitlab.com/Monsterovich/fixedstring` URL of your library to the `repositories.txt` file, your pull request now proposes to remove the URL of another person's library (`https://github.com/fatlab101/FixedString`) from the file. This is of course not acceptable.

You can see the problem if you look at the "Files changed" tab of this pull request:

https://github.com/arduino/library-registry/pull/9077/changes

(note the line highlighted in pink)

Please correct your pull request so that it only adds the URL of your library, and doesn't make any other modifications to the `repositories.txt` file.

Monsterovich:
> This is of course not acceptable.

@per1234  If I revert it, the bot will complain that there are two repositories with the same name. Besides, that implementation is far inferior.

per1234:
>the bot will complain that there are two repositories with the same name.

Obviously you must fix that by changing the name of your library, not by taking the unilateral decision to remove someone else's library.

How would you like it if we allowed some random person to make the decision your library was "inferior" and should be removed without your permission???

Monsterovich:
> Obviously you must fix that by changing the name of your library

Figure it out on your own; it's all your responsibility.

> How would you like it if we allowed some random person to make the decision your library was "inferior" and should be removed without your permission???

I don't care. However, collecting a bunch of libraries that all serve the same purpose and turning the repository into a mess - that sounds like a great idea!

PR #9132, opened 2026-09-16
EEPROM24CXX

AlenKrishna2012:


per1234 (commented):
Closing as duplicate of https://github.com/arduino/library-registry/pull/8876.

@AlenKrishna2012 please don't submit pointless pull requests to this repository.

---

AlenKrishna2012:
I tryed to update the library but info always shows that the "release 1.0.x already loaded,skipping" but it never appear in the official library manager web page. I did everything like changing library.properties file and added new tag with release but its saying the same thing so that is why i am trying to retry every step again. Can you please help to do this?

per1234:
>it never appear in the official library manager web page

By "official library manager web page" are you referring to the Arduino Library Reference? I see that there is indeed a reference page for your library:

https://docs.arduino.cc/libraries/eeprom24cxx/

AlenKrishna2012:
Oh! it was never there in the in the first week so i stoped looking for it and after that i try to restart again. But why it always show "release x.x.x already loaded, skipping x.x x" why is that?

per1234:
Arduino Library Manager uses data provided by an index file hosted on Arduino's server. This index file is updated by a special tool. We run this tool periodically using a scheduled job. On each run of the tool, it goes through the list of registered libraries. For each library, the tool examines each of the [tags](https://git-scm.com/docs/git-tag) in the library repository. If it finds there is already an entry in the existing index file for a given tag, it skips any further processing of that tag. The message like "`Release EEPROM24CXX:1.0.7 already loaded, skipping`" is printed in this case.

So the occurrence of these messages in the indexer job run logs for your library is normal and expected. They tell you that the specified release has been successfully added to the Arduino Library Manager index.

The primary thing that would be of concern in logs would be an error message produced by the tool when processing a release tag. Another thing to look out for is the case where you see some unexpected version number for a tag. Here is a hypothetical example of what I am referring to:

```text
2026/09/18 17:36:31 Checking out tag: 1.0.7
2026/09/18 17:36:31 Release EEPROM24CXX:1.0.7 already loaded, skipping
2026/09/18 17:38:01 Checking out tag: 1.0.8
2026/09/18 17:38:01 Release EEPROM24CXX:1.0.7 already loaded, skipping
```

This would indicate that you forgot to update the value of the `version` property in your library's `library.properties` file before making the `1.0.8` tag in the library repository.

When we look at the logs for your library, we only see that the 1.0.7 release was successfully indexed. There are no errors or signs of any other problem with the library. So everything is perfectly fine and you should not worry any more about your library.

AlenKrishna2012:
Thanks for your help of making me understand the log and giving the link to the repository! I would enquire if there is any problem.

PR #9154, opened 2026-09-19
Update Preflight-Arduino URL after org transfer

horse-pants:
The repository has moved from `horse-pants/Preflight-Arduino` to [`diydoohickeys/Preflight-Arduino`](https://github.com/diydoohickeys/Preflight-Arduino) — same library, transferred to an organization. Submitted originally in #9153.

The old URL still redirects, but updating the index entry to point at the canonical location.


horse-pants:
To clarify for the reviewing maintainer: **this is an intentional URL change**, per [maintenance-requests.md](https://github.com/arduino/library-registry/blob/main/maintenance-requests.md#update-library-url). The library name is unchanged — it is `Preflight` at both the old and new URL.

The `LP018` error looks like a timing artifact rather than a real problem. The library was only added a couple of hours ago in #9153, and the indexer has not run yet (its log at `downloads.arduino.cc/libraries/logs/github.com/horse-pants/Preflight-Arduino/` is still 404), so `Preflight` is not yet present in the index for the check to find. Nothing has been renamed.

The repository was transferred to an organization; the old URL still redirects, so the entry works either way — this just points the index at the canonical location.

per1234 (commented):
Hello @horse-pants. I am confirming your request.

Information about that here:

PR #9175, opened 2026-09-22
Add esp-rtosLogger repository link

HamzaYslmn:
Async Serial Writer


HamzaYslmn:
Tag 0.0.1 is now pushed and matches the version in library.properties.

@ArduinoBot

